PR #4, opened 2018-06-07
Add examples

bryanwweber:
Add the Python, Matlab, and Jupyter examples as rendered files


bryanwweber:
Unfortunately, the code to generate the examples has to be duplicated quite a bit because functions don't get called unless they're called by some internal mechanism that I don't really understand at this point. So the code in `build_examples.py` could probably be much better, I just don't know how 😄

bryanwweber:
This PR is GTG :+1: on my end now!

PR #5, opened 2018-06-09
Add error pages

bryanwweber:
403, 404, 500, 503 error pages are produced in the root of the site.
Need to configure the web server to serve them when an error occurs.

### DIFF
--- a/pages/403.html
+++ b/pages/403.html
@@ -1,0 +1,13 @@
+<!--
+  .. title: 403 Forbidden
+  .. slug: 403
+  .. errdesc: Page Forbidden
+  .. date: 1970-01-01
+  .. description: You're not allowed to access that page.
+  .. pretty_url: False
+  .. template: error.tmpl
+-->
+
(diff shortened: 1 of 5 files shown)


bryanwweber:
https://httpd.apache.org/docs/2.4/custom-error.html

pages/403.html
```diff
@@ -0,0 +1,13 @@
...
+  .. title: 403 Forbidden
+  .. slug: 403
+  .. errdesc: Page Forbidden
+  .. date: 1970-01-01
+  .. description: You're not allowed to access that page.
+  .. pretty_url: False
+  .. template: error.tmpl
+-->
+
+<p>Please check the URL. If you clicked a link and ended up here, please
```
speth: I would say "...a link **on cantera.org** and ended up here..." to avoid encouraging people to tell us about dead links elsewhere on the internet.

bryanwweber: Good point!

PR #6, opened 2018-06-10
Add Python and Matlab tutorials

decaluwe:
Python and Matlab tutorials added for the new website.

- Python tutorial is much the same as before, except some of the content about working with cti files is removed, and is replace by a link to content elsewhere in the documentation.
- Matlab tutorial content has been translated from the 'tutX.m' files to be displayed on the website, similar to the Python tutorial.  I think it is still wise to also keep the m-files so they can also be run locally by the user, if desired.
- TO DO: Internal references/hyperlinks currently just link to the top of the page, but it would be good to find a way to link directly to headings/sub-headings/etc.


pages/tutorials/index.html
```diff
@@ -33,52 +37,58 @@ <h1 class="display-4">Getting started with Cantera</h1>
...
-      <li> Information on how to locate and work with Cantera input files (which contain the thermodynamic, transport,
-        and chemical kinetic information for the phases of interest).
+      <li> Information on how to locate and work with Cantera input files (which
+        contain the thermodynamic, transport, and chemical kinetic information
+        for the phases of interest).
       </li>
-      <li> Detailed documentation ans user guides for accessing Cantera via Python, Matlab, and directly
-        via C++. For advanced and intermediate users, the documentation is an easily-searchable repository
-        for information on specific functions of interest.
+      <li> Detailed documentation ans user guides for accessing Cantera via
```
bryanwweber: Small typo here `ans`->`and`

pages/tutorials/index.html
```diff
@@ -107,7 +117,7 @@ <h1 class="card-title">Users' Guides.</h1>
         <div class="card-footer">
           <a href="../cython/index.html" class="btn btn-primary">Python </a>
```
bryanwweber: This link is going to be `../documentation/index.html#python` and the Matlab one similarly

pages/tutorials/index.html
```diff
@@ -107,7 +117,7 @@ <h1 class="card-title">Users' Guides.</h1>
...
           <a href="../matlab/index.html" class="btn btn-primary">Matlab </a>
-          <a href="../matlab/cxx.html" class="btn btn-primary">C++ </a>
+          <a href="../cxx/index.html" class="btn btn-primary">C++ </a>
```
bryanwweber: I'm not sure exactly how this link will turn out yet, so we can update it later

pages/tutorials/index.html
```diff
@@ -107,7 +117,7 @@ <h1 class="card-title">Users' Guides.</h1>
...
         </div>
       </div>
       <div class="card">
```
bryanwweber: The examples link turned out to be `../examples/index.html`

pages/tutorials/input-files.rst
```diff
@@ -0,0 +1,716 @@
+.. py:currentmodule:: cantera.ctml_writer
```
bryanwweber: This line, and line 3, are Sphinx-specific things that I think we should probably drop

pages/tutorials/index.html
```diff
@@ -33,52 +37,58 @@ <h1 class="display-4">Getting started with Cantera</h1>
...
         <div class="card-body">
-          <h1 class="card-title">I want to learn about Cantera via the Matlab toolbox.</h1>
+          <h1 class="card-title">I want to learn about Cantera via the Matlab
+            toolbox.</h1>
         </div>
         <div class="card-footer">
-          <a href="../matlab/tutorial.html" class="btn btn-primary">Matlab Tutorial</a>
+          <a href="MatlabTutorial.html" class="btn btn-primary">Matlab Tutorial</a>
```
bryanwweber: The links here need to be to the slug, i.e., `matlab-tutorial.html`, except your slug line isn't right, it should be `.. slug: matlab-tutorial` (only one `:` and a space before the `matlab-tutorial`. Same for Python. If you'd rather keep it this way, URLs are case-insensitive, so these end up being all lower-case file names (at least on my machine) although everything works out because of the case-insensitivity.

pages/tutorials/index.html
```diff
@@ -33,52 +37,58 @@ <h1 class="display-4">Getting started with Cantera</h1>
...
+        information on specific functions of interest.
       </li>
-      <li> A repository of examples, demonstrating how to use Cantera to solve a diverse range of problems.
-        You can either use these examples directly, or use them as a template to develop your own applications.
+      <li> A repository of examples, demonstrating how to use Cantera to solve a
+        diverse range of problems. You can either use these examples directly,
+        or use them as a template to develop your own applications.
       </li>
     </ul>
   </p>
```
bryanwweber: I can't comment on the line for `InputFiles`, but for that one you have the slug command set properly, so the file is `input-tutorials.html`, and you'll want to drop the leading `/`, which points to a file in the root of the website.

pages/tutorials/index.html
bryanwweber: Alternatively, making it relative to the site root would probably be better: `/documentation/index.html`

pages/tutorials/index.html
decaluwe: Okay, I think I got these cleaned up.

PR #7, opened 2018-06-12
[WIP] Beginning CTI tutorial

decaluwe:
Establishes landing/intro page for tutorials on working with cti files. Still getting some of the elements in place.

- Leaving errors in place, for now, relating to (what I think are) sphinx-related tags like `:ref:` and `:class:`.
- I currently have it set up as a series of modules, which can be processed in series, or selectively as needed.
- Would probably be a good idea to have navigation (`next`, `previous`, `home`) buttons incorporated.


bryanwweber:
The `:func:` role links to a function from the API documentation. These sections should be carefully considered as to what belongs here and what should be transferred over the API section (via updates to the main Cantera repository).

It would probably be easiest to hard-code the buttons, rather than relying on templating them.

PR #29, opened 2018-07-31
Docs rename script

bryanwweber:
Renames the duplicate HTML files created by Doxygen in older versions of the Cantera docs and rewrites all the links to these files.


bryanwweber:
I'm leaving the history of the Python script that did the conversion so that it can be resurrected in the future if necessary.

Issue #42, opened 2018-10-09
The file http://www.cantera.org/license.txt doesn't exist


bryanwweber:
This URL is linked from all our header files, so the link should probably work...

bryanwweber:
The file can be found at https://github.com/Cantera/cantera/blob/master/License.txt and should be copied to the root of the deployed website.

PR #44, opened 2018-10-11
Add license.txt

petrkle:
Add license.txt on web #42


bryanwweber:
Hi! 👋 Thanks for making this pull request. I think it would be better to explicitly list all the files to be copied, rather than relying on copying the whole folder.

bryanwweber:
Hi @petrkle! Do you plan to update this pull request for the change I mentioned?

petrkle:
Sorry, I'am not so familiar with Nikola. If I try copy one file (not directory) it end with error. You can close this PR.

bryanwweber:
@petrkle Right, so this turns out to be a bit more complicated than I thought, plus we're going to be adding more files, so I'm just going to merge this. Thank you for your contribution!

PR #86, opened 2019-08-08
Force cantera version 2.4

ischoegl:
Force checkout of cantera 2.4 to exclude examples only running on cantera 2.5.0a3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
     - cantera.org
 
 before_install:
-  - git clone --depth=1 https://github.com/Cantera/cantera ../cantera
+  - git clone --depth=1 https://github.com/Cantera/cantera --branch 2.4 --single-branch ../cantera
   - git clone --depth=1 https://github.com/Cantera/cantera-jupyter ../cantera-jupyter
 
 script:


speth:
I see your point, but if we're going to do this then we need another version of the website that applies to the development version, which means maintaining two branches of the website repo. I'm not saying we shouldn't do this, but it is going to create more maintenance work.

ischoegl:
@speth ... this follows a suggestion by @bryanwweber. 

I agree that there may be a point in differentiating docs that do not apply to the stable version from what is only in the development version (e.g. the updated `FlowDevice` science section). There are numerous ways to approach this but probably all involve more maintenance ... this is obviously not my decision to make. Bryan's approach is certainly the most efficient to keep development code from causing trouble (happened to me while testing jupyter notebooks meant for 2.4 prior to the release).

bryanwweber:
This change would apply specifically and only to the examples. I think it's probably best to keep examples consistent with the most recent release, since old examples should continue to work with the dev version but not vice versa

speth:
But why should this only apply to the examples, and not to changes like those that @ischoegl just made to `FlowDevice`, or the new documentation I still need to write for the YAML format, which only apply to the next release?

In any case, I think we need a version of the website living somewhere which is connected to the development version. Otherwise, we'll end up in the situation where say, a new category of examples is added that breaks the web site in some way which isn't discovered until we update for the release.

ischoegl:
I am not completely aware of all the intricate details of the build process, but there may be a relatively simple approach to allow for both `cantera` and `cantera-devel` material to coexist. 

If you were to move the undifferentiated (but version-affected) material on *Cantera/cantera-website* to a new repo that has branches (e.g. *Cantera/cantera-docs*), it should be possible to automatically generate two separate versions by pulling each repo twice (including documentation, examples, jupyter and science pages): one would be `cantera` (latest stable branch), the other `cantera-devel` (master branch). Just my 2 cents ...

bryanwweber:
@speth Yes, I agree we need to organize that new material somehow. However, this change specifically is only for the examples, which I think should be handled separately for the reason I mentioned before.

For new content that only applies to the development edition, I think that should be handled in a separate PR and requires some more thought.

One option is that we can add all new materials to `master`, and have a special box or other mechanism to note that the material only applies to an unreleased version of Cantera. I'm not exactly sure how this would work, and it would certainly require quite a bit of maintenance to make sure that when a new release happened, everything was updated properly, but this might be able to be automated, and could serve as a neat flag on the website ("added in 2.5.0").

A second option is to have a separate branch for new content (`dev` or something) that gets merged into `master` when a new release happens and is basically a staging area for new content. This would be mildly more difficult to show to the public, because it would be a whole copy of the website hosted at a different URL, but might result in a little bit less maintenance than having to remove all the version markers.

@ischoegl In either of the cases I mentioned above, I don't think we'd need a separate repo. The API docs (i.e., those generated by Sphinx and Doxygen from the source-code docstrings) are already uploaded to the website at https://cantera.org/documentation/dev-docs.html The only change we need to make is when we want to add new documentation to the other pages on the website, like adding something to the Science about the new P-R EOS or your `FlowDevice` changes.

Maybe what we should do is see try to see what some of the other NumFOCUS projects do in terms of how public they make their dev documentation and get a sense for how to proceed?

PR #94, opened 2019-10-26
Suggest installing jupyter in new conda enviroment

rwest:
If you don't install jupyter in the `spam` conda enviroment you create for using cantera, then when you try to launch jupyter your shell will find some other jupyter and launch that one instead without warning, and you'll be confused that _even though you have clearly activated your `spam` environment_ you still can't import cantera.   As we trend towards using jupyter notebooks as demos, I see this becoming a more frequent mistake.

(One could consider removing ipython from the list at some point. I presume this was put there for a similar reason)

### DIFF
--- a/pages/install/conda-install.rst
+++ b/pages/install/conda-install.rst
@@ -13,8 +13,8 @@
 
     .. class:: lead
 
-      `Anaconda <https://www.anaconda.com/downloads>`__ and
-      `Miniconda <https://conda.io/miniconda.html>`__ are Python distributions that include the
+      `Anaconda <https://www.anaconda.com/products/individual#Downloads>`__ and
+      `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`__ are Python distributions that include the
       ``conda`` package manager, which can be used to install Cantera. Both distributions are
       available for Linux, macOS, and Windows. Note that installing Cantera using Conda will only
       provide the Cantera Python module. If you want to use the other Cantera interfaces
@@ -25,16 +25,18 @@ Anaconda and Miniconda both include ``conda``; the difference is that Anaconda i
 number of Python packages that are widely used in scientific applications, while Miniconda is a
 minimal distribution that only includes Python and Conda, although all of the packages available in
 Anaconda can be installed in Miniconda. For more details on how to use conda, see the `conda
-documentation <https://conda.io/docs/intro.html>`__.
+documentation <https://docs.conda.io/projects/conda/en/latest/user-guide/index.html>`__.
 
 **Option 1: Create a new environment for Cantera**
 
 If you have just installed Anaconda or Miniconda, the following instructions
 will create a conda environment where you can use Cantera. For this example, the
-environment is named ``spam``. From the command line (or the Anaconda Prompt
-on Windows), run::
+environment is named ``ct-env``. From the command line (or the Anaconda Prompt
+on Windows), run:
 
-    conda create --name spam --channel cantera cantera ipython matplotlib jupyter
+.. code:: shell
+
+   conda create --name ct-env --channel cantera cantera ipython matplotlib jupyter
 
 This will create an environment with Cantera, IPython, Matplotlib, and all their
 dependencies installed. Although Conda can install a large set of packages by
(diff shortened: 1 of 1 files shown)


bryanwweber:
The official IPython docs suggest installing `ipykernel` instead, but it requires a few more steps: https://ipython.readthedocs.io/en/stable/install/kernel_install.html#kernels-for-different-environments Any thoughts there?

ischoegl:
As another suggestion, it would be simple to ship a *recommended* `environment.yml` with cantera, which takes care of everything (including `graphviz` and `python-graphviz` required for `ct.ReactionPathDiagram`) via:
```
conda env create -f environment.yml
```

bryanwweber:
We could certainly suggest something on the website here

ischoegl (approved):
While I'm approving the PR as is (I believe it's an improvement), my earlier comment about `environment.yml` is still valid. Adapted from my own work, I'd recommend:
```yaml
# cantera conda package dependencies
name: ct-env
channels:
  - cantera # replace by 'cantera/label/dev' for development version
  - defaults
dependencies:
  - python=3.8 # version 3.6 or later is recommended
  - numpy
  - ruamel.yaml
  - cantera # installs latest stable version
  - ipython
  - jupyter
  - matplotlib
  - python-graphviz # optional (needed for reaction path diagrams)
  - h5py # optional (needed for HDF/H5 output)
  - pandas # optional (needed for pandas interface)
```
This obviously would be followed by:
```
$ conda env create -f environment.yml
$ conda activate ct-env
```

**PS:** I had another comment about the usage of *spam* that was lost. While the origins are obvious to many, it may confuse others. It may make sense to replace by `ct-env` here and `my-env` below.

speth:
> I had another comment about the usage of _spam_ that was lost. While the origins are obvious to many, it may confuse others.

I must admit I am a little surprised by the number of users that have indeed gone ahead and created conda environments named `spam`. The intention when I wrote those was that the sort of off-the-wall names (whether or not one catches the Monty Python reference) would make it more obvious that those parts of the command could be replaced with whatever you wanted. I guess, with something like `ct-env`, I would imagine that more people will use the provided name directly, which will lead to less bizarre usage, if not necessarily better-informed.

I'm 👍 on the idea of providing a template `environment.yml` file on the website.

decaluwe:
I'll just provide a trivial comment, that `ct-env` is indeed easy to type; and therefore I like it.  Also, as someone who has literally typed `/path/to/conda/install/folder/bin/conda` into a command line, I am also still somehow surprised how often people create an `env` named `spam` 😂 

Regarding @ischoegl 's recommendation, just to spell it out for us dummies: the procedure would be to download `environment.yaml` from the website, `cd` to wherever it is saved, then run the `conda env create...` command from his most recent post?

ischoegl:
I believe from what I hear the consensus would probably be to add a code block with the content of `environment.yml` on the website, and instruct users to paste that into a new file based on past experience, cut&paste should work well). As YAML is human-readable, it would serve as documentation itself.

bryanwweber:
@ischoegl @decaluwe @speth I added some instructions for using the `environment.yaml` file, let me know what you think.

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
+This option is similar to **Option 1** but includes a few other packages that
+you may find helpful as you're working with Cantera. Copy and paste the contents
+of the file shown below into a file called ``environment.yaml``. Then, save the
+the file somewhere and remember that location.
+
+.. code:: yaml
+
+   name: ct-env
+   channels:
+   - cantera
```
ischoegl, line 68: I'd still leave the comment about the development version in - it's an opportunity to document the existence of the 'bleeding edge' package (and the last bullet can reference back).

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
+   - defaults
+   dependencies:
+   - python
```
ischoegl, line 71: Similarly, pointing out that python 3.6 and up (with default option) may be an opportunity for documentation.

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
+need to activate the environment before using any scripts or modules that you
+just installed:
+
+.. code:: shell
+
+   conda activate ct-env
+
+**Option 3: Install Cantera in an existing environment**
+
+First, activate your environment (assumed here to be named ``ct-env``; if you've
```
ischoegl, line 104: perhaps `my-env` to differentiate from above?

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
 
 If you later want to revert back to the stable version, first remove and then
-reinstall Cantera::
+reinstall Cantera:
+
+.. code:: shell
 
-    conda remove cantera
-    conda install --channel cantera cantera
+   conda remove cantera
```
ischoegl, line 132: I personally use `conda remove --name ct-dev --all`

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
 
-    conda install --channel cantera/label/dev cantera
+**Option 4: Install the development version of Cantera**
+
+To install a recent development snapshot (that is, an alpha or beta version) of
+Cantera in an existing environment, activate the environment and then run:
+
+.. code:: shell
+
+   conda install --channel cantera/label/dev cantera
```
ischoegl, line 125: `ct-dev` to differentiate again? If the comments in `environment.yaml` could be left in, it would dove-tail nicely.

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
+
+This will create an environment called ``ct-env``. Once you've done that, you
+need to activate the environment before using any scripts or modules that you
+just installed:
+
+.. code:: shell
+
+   conda activate ct-env
+
+**Option 3: Install Cantera in an existing environment**
```
ischoegl, line 102: Do we want to encourage this? There is no real benefit for people familiar with `conda` (`environment.yml. already contains all dependencies that need to be known), while others may get tangled up quite a bit.

pages/install/conda-install.rst
```diff
@@ -45,32 +47,87 @@ obtained are specified by adding the ``--channel`` option in the ``install`` or
...
+   name: ct-env
+   channels:
+   - cantera
+   - defaults
+   dependencies:
+   - python
+   - cantera
+   - ipython  # optional (needed for nicer interactive command line)
+   - jupyter  # optional (needed for Jupyter Notebook)
+   - matplotlib  # optional (needed for plots)
```
ischoegl, line 75: `numpy` and `ruamel.yaml` are missing.

ischoegl (commented):
@bryanwweber ... nice! Thanks for taking this on. Have a few minor comments ...

pages/install/conda-install.rst
bryanwweber: I'd rather that people install the most recent version of Python that we support, and not specify a particular version here. IMO, the more we can push people to quickly upgrade versions of Python, the more features from recent versions we can use. I can add a comment like:

```yaml
- python  # Cantera supports Python 3.5 and up
```

which is true for 2.5.1.

pages/install/conda-install.rst
bryanwweber: They are automatically brought in with Cantera. I believe the conda recipes depend on `ruamel_yaml`, not `ruamel.yaml` (IIRC due to some namespace packaging issues in conda. There was discussion between Anton and the conda devs a few years ago leading to the different packages yadda yadda yadda). We end up being able to use either with all the `try..except ImportError`, but there's no need to double-specify, IMO.

pages/install/conda-install.rst
bryanwweber: That's a good point, I think we can remove this option.

pages/install/conda-install.rst
ischoegl: I'd still 'recommend' 3.6 and up, but that's minor.

pages/install/conda-install.rst
bryanwweber: This is explicitly to replace the dev version with a stable version in an existing environment where you've already installed the dev version. I can clarify that the environment should be activated.

bryanwweber:
One thing I forgot to mention is that we should document the `conda-forge` package, particularly since that's the only one available for macOS ARM. I started doing that here, then figured it should be a separate PR, because I also wanted to add installing the MATLAB and C++ interfaces, which can now be done with Conda.

pages/install/conda-install.rst
bryanwweber: I take that comment back, we don't have conda packages for Python 3.5, see https://anaconda.org/Cantera/cantera/files?version=2.5.1 So I'll change that to say "3.6 and up"

pages/install/conda-install.rst
ischoegl: Rather than replacing the cantera version, it may make sense to recommend creating a separate one. Installation time isn't too much of an issue, and unused environments can be simply removed if space becomes an issue. I never found it worth the time to 'maintain' an environment, as re-creating things is usually much quicker.

bryanwweber: I left a note that creating a new environment is also an option. I want to leave the instructions for how to replace a package in an existing environment, but I also clarified the environment in use.

ischoegl (approved):

bryanwweber:
Thanks everyone!

Issue #17, opened 2018-07-01
Add copybutton functionality similar to Python docs


bryanwweber:
See the js file here: https://docs.python.org/3/_static/copybutton.js

bryanwweber:
Closed by #98! :tada:

PR #98, opened 2019-12-08
Add copybutton functionality

paulblum:
Closes issue #17.

This pull request adds the copybutton functionality from Python Docs to all Python Console and Matlab Command Window codeblocks on the Cantera website.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -1176,11 +1176,7 @@ USE_CDN_WARNING = False
 # Extra things you want in the pages HEAD tag. This will be added right
 # before </head>
 # (translatable)
-EXTRA_HEAD_DATA = """
-<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
-<script src="/assets/js/copybutton.js"></script>
-"""
-
+# EXTRA_HEAD_DATA = ""
 # Google Analytics or whatever else you use. Added to the bottom of <body>
 # in the default template (base.tmpl).
 # (translatable)
--- a/conf.py
+++ b/conf.py
@@ -1176,7 +1176,11 @@ USE_CDN_WARNING = False
 # Extra things you want in the pages HEAD tag. This will be added right
 # before </head>
 # (translatable)
-# EXTRA_HEAD_DATA = ""
+EXTRA_HEAD_DATA = """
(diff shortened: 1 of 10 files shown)


conf.py
```diff
@@ -1163,7 +1163,11 @@
 # Extra things you want in the pages HEAD tag. This will be added right
 # before </head>
 # (translatable)
-# EXTRA_HEAD_DATA = ""
+EXTRA_HEAD_DATA = """
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
```
bryanwweber: JQuery should already be loaded at some point on the page, likely at the bottom. You might need to include the `copybutton.js` file at the bottom of the page instead of the top to fix any errors that occur.

plugins/override_html_formatter.py
```diff
@@ -0,0 +1,25 @@
...
+    """Override Nikola's modification of the Pygments HTML Formatter,
+    reverting it back to the unmodified version."""
+
+    name = "override_html_formatter"
+
+    def set_site(self, site):
+        """Set the Nikola site."""
+        self.site = site
+        nikola.utils.NikolaPygmentsHTML = PygmentsHtmlFormatter
+        return super(OverrideHTMLFormatter, self).set_site(site)
```
bryanwweber: In Python 3, you can drop the arguments to `super()`.

themes/cantera/assets/js/copybutton.js
```diff
@@ -0,0 +1,60 @@
+$(document).ready(function () {
```
bryanwweber: If this is copy-pasted from the Python docs, we probably need to include the license somewhere in it.

bryanwweber: https://github.com/python/python-docs-theme/blob/master/python_docs_theme/static/copybutton.js

bryanwweber:
So for the MATLAB sessions, there shouldn't need to be a prompt set on the second line after a continuation. This is a bug in Pygments and I filed Pygments/pygments#1399 to address this. I just want to see if I can get a response there before merging this.

paulblum:
@bryanwweber I noticed that too, added the prompts so that the continuation wouldn't erroneously be displayed as output, as was the case with the bug. Thanks for fixing it!

Issue #88, opened 2019-08-13
How should we handle updates to website content for the development version?


bryanwweber:
As discussed in #23, #86, Cantera/cantera#667, and Cantera/cantera#682, certain updates to Cantera/cantera will require updates to the content on the website. The question is, how should that content be shown? The options that I think have been discussed so far:

1) A separate branch in this repository to house the updates that only apply to the development edition. This could work like it does at Cantera/cantera, where the main/"default" website is stored in a branch named for the current release and ongoing work goes onto `master`, or `master` can remain the default version and new updates go onto a `dev` branch. This would probably result in two separate versions of the website deployed at two URLs (e.g., `https://cantera.org` and `https://devel.cantera.org` or some such).
2) Add all new materials to `master`, and have a special box or other mechanism to note that the material only applies to an unreleased version of Cantera. I'm not exactly sure how this would work, and it would certainly require quite a bit of maintenance to make sure that when a new release happened, everything was updated properly, but this might be able to be automated, and could serve as a neat flag on the website ("added in 2.5.0").
3) Store the related materials in a separate repository, with content on two branches (one for the current release, one for the development version). This would minimize the amount of duplication in building the website (i.e., option 1 results in two whole websites), but would require a bit more CI wrangling and would make building the website locally more annoying.

This issue refers specifically to the science, equations, supported models, and tutorials. API developer documentation is [already available](https://cantera.org/documentation/dev-docs.html).

In my opinion, examples should always be shown from the current released version of Cantera, since that content should continue to work in the developer version of Cantera, while the converse is not true; in addition, advanced users who're using the development version should be able to get updated examples from their copy of the source.

A related point that I want to quote here specifically from @speth (https://github.com/Cantera/cantera-website/pull/86#issuecomment-519663694):

> [We should try to avoid the case where] a new category of examples is added that breaks the web site in some way which isn't discovered until we update for the release.

To that point, examples that only use existing features (in the current stable release) should be added to the release branch in Cantera/cantera, and if this involves new categories, that should be able to be shown on the website.

ischoegl:
@bryanwweber ... any update on this? At some point, I should probably document what was implemented in Cantera/cantera#667 . I guess it can be 'parked' in a pull request.

bryanwweber:
@ischoegl Yes, parking it in a pull request is probably the best scenario for now

ischoegl:
While there is the potential of merge conflicts, I believe is should be possible to merge changes into the `testing` branch?

[...]

PR #104, opened 2020-05-16
[Cantera 2.5] Update valve/etc. in science section

ischoegl:
Addresses @speth's comment in Cantera/cantera#667

> The one remaining to-do for this change is to update the "science" documentation for the reactor model on cantera.org (https://cantera.org/science/reactors.html#reactor-networks-and-devices).

Per @bryanwweber's comment in #88, the update should be parked as a PR until the release of 2.5.

### DIFF
--- a/pages/science/reactors.rst
+++ b/pages/science/reactors.rst
@@ -326,26 +326,27 @@ to the reactors previously mentioned:
 
   .. math::
 
-     \dot m = c g(t) f(\Delta P)
+     \dot m = K_v g(t) f(P_1 - P_2)
 
-  with *c* being a proportionality constant. Further, *g* and *f* are functions
-  of time :math:`t` and pressure drop :math:`\Delta P` that are set by class
-  methods :py:meth:`setTimeFunction` and :py:meth:`setPressureFunction`,
+  with :math:`K_v` being a proportionality constant that is set using the class
+  property :py:attr:`Valve.valve_coeff`. Further, :math:`g` and :math:`f`
+  are functions of time and pressure drop that are set by class methods
+  :py:func:`Valve.set_time_function` and :py:func:`Valve.set_valve_function`,
   respectively. If no functions are specified, the mass flow rate defaults to:
 
   .. math::
 
-     \dot m = c \Delta P
+     \dot m = K_v (P_1 - P_2)
 
   The pressure difference between upstream (*1*) and downstream (*2*) reservoir
-  is defined as :math:`\Delta P = P_1 - P_2`. It is never possible for the
-  flow to reverse and go from the downstream to the upstream reactor/reservoir
-  through a line containing a Valve object, which means that the flow rate is
-  set to zero if :math:`P_1 < P_2`.
+  is defined as :math:`P_1 - P_2`. It is never possible for the flow to reverse
+  and go from the downstream to the upstream reactor/reservoir through a line
+  containing a Valve object, which means that the flow rate is set to zero if
+  :math:`P_1 < P_2`.
 
   Valve objects are often used between an upstream reactor and a downstream
   reactor or reservoir to maintain them both at nearly the same pressure. By
-  setting the constant :math:`c` to a sufficiently large value, very small
+  setting the constant :math:`K_v` to a sufficiently large value, very small
   pressure differences will result in flow between the reactors that counteracts
   the pressure difference.
 
@@ -355,15 +356,16 @@ to the reactors previously mentioned:
 
   .. math::
 
-     \dot m = m g(t)
+     \dot m = m_0 g(t)
 
-  where *m* is a mass flow coefficient and *g* is a function of time that is set
-  by the class method :py:meth:`setTimeFunction`. If no function is specified,
-  the mass flow rate defaults to:
+  where :math:`m_0` is a mass flow coefficient and :math:`g` is a function of time
+  which are set by class property :py:attr:`MassFlowController.mass_flow_coeff`
+  and method :py:func:`MassFlowController.set_time_function`, respectively. If no
+  function is specified, the mass flow rate defaults to:
 
   .. math::
 
-     \dot m = m
+     \dot m = m_0
 
   Note that if :math:`\dot m < 0`, the mass flow rate will be set to zero,
   since a reversal of the flow direction is not allowed.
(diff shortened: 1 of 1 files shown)


bryanwweber:
@ischoegl You can rebase this onto `master` now

bryanwweber:
@ischoegl What's the status of this PR? Thanks!

ischoegl:
@bryanwweber ... will have to look over recent changes on the MFC. Will certainly tackle this as soon as beta is reached.

bryanwweber:
Thanks @ischoegl I agree waiting for beta makes sense.

ischoegl:
@bryanwweber ... now that 2.5 is on the horizon, I finally updated the docs.

pages/science/reactors.rst
```diff
@@ -322,26 +322,30 @@ to the reactors previously mentioned:
   detail of how the wall affects the connected reactors.
 
 - :py:class:`Valve`: A valve is a flow devices with mass flow rate that is a function of
-  the pressure drop across it. The default behavior is linear:
+  the pressure drop across it. The mass flow rate is computed as:
 
   .. math::
 
-     \dot m = K_v (P_1 - P_2)
+     \dot m = c g(t) f(\Delta P)
```
bryanwweber, line 329: I would prefer to keep this explicitly as `P_1 - P_2`, since `\Delta` usually implies `2 - 1`. I see you have the definition later, but it's still a little counter intuitive to me.

pages/science/reactors.rst
```diff
@@ -322,26 +322,30 @@ to the reactors previously mentioned:
...
 
-  if :math:`P_1 > P_2.` Otherwise, :math:`\dot m = 0`. However, an arbitrary
-  function can also be specified, such that
+  with *c* being a proportionality constant. Further, *g* and *f* are functions
```
bryanwweber, line 331: Can you put `c`, `g`, and `f` in math formatting here?

pages/science/reactors.rst
```diff
@@ -351,11 +355,18 @@ to the reactors previously mentioned:
 
   .. math::
 
-     \dot m = \max(\dot m_0, 0.0)
+     \dot m = m g(t)
 
-  where :math:`\dot m_0` is either a constant value or a function of time. Note
-  that if :math:`\dot m_0 < 0`, the mass flow rate will be set to zero, since
-  reversal of the flow direction is not allowed.
+  where *m* is a mass flow coefficient and *g* is a function of time that is set
```
bryanwweber, line 360: Can you put `m` and `g` in math formatting? I'm also not sold on the symbol `m` for the coefficient. It would probably be good to use `\dot{m}_0` like in the class docstring. Can you specify how one sets this coefficient, like you've done for the function?

pages/science/reactors.rst
```diff
@@ -374,7 +385,19 @@ to the reactors previously mentioned:
 
   .. math::
 
-     \dot m = \dot m_{\mathrm{master}} + K_v(P_1 - P_2).
+     dot m = \dot m_{\mathrm{master}} + c f(\Delta P)
```
bryanwweber, line 388:
```suggestion
     \dot m = \dot m_{\text{master}} + c f(P_1 - P_2)
```

Same suggestion here about `P_1 - P_2`, plus a typo fix

pages/science/reactors.rst
```diff
@@ -374,7 +385,19 @@ to the reactors previously mentioned:
...
+
+  where *c* is a proportionality constant and *f* is a function of pressure drop
```
bryanwweber, line 390: Same comment here about math mode for `c` and `f`.

pages/science/reactors.rst
```diff
@@ -374,7 +385,19 @@ to the reactors previously mentioned:
...
+  :math:`\Delta P = P_1 - P_2` that is set by the class method
+  :py:meth:`setPressureFunction`. If no function is specified, the mass flow rate
+  defaults to:
+
+  .. math::
+
+     \dot m = \dot m_{\mathrm{master}} + c \Delta P
```
bryanwweber, line 397:
```suggestion
     \dot m = \dot m_{\text{master}} + c \Delta P
```

bryanwweber (changes requested):
A few formatting comments here. I think it would make more sense to keep the symbols here consistent with the docstrings, keeping `K_v` instead of `c`.

One problem is that the `:py:meth:` role isn't defined, but I can't recall right now how to link to Python documentation methods... I think it's `:py:func:` with the fully classified name (i.e., `MassFlowController.set_time_function`). You can shorten the name to just the method by putting a tilde before the class name.

That said, there doesn't actually appear to be any documentation for the various attributes and methods of these classes, see: https://cantera.org/documentation/dev/sphinx/html/cython/zerodim.html#massflowcontroller I'm not sure what happened to those.

pages/science/reactors.rst
ischoegl: True. I'm following code here, which uses the opposite sign - this should probably be changed to be in line with conventions (I *think* this choice predates my updates). As an aside, the parameter names have been changed in the docstrings also, as they have a more generic meaning after the code change (i.e. coefficients are now defined as `m_coeff` in the `FlowDevice` class).

pages/science/reactors.rst
ischoegl: Code changed, as have the class docstrings.

ischoegl:
@bryanwweber ... done. I'm not reverting symbols you mentioned, as I am following current docstrings. Regarding pressure drop, it may make sense to use the usual convention in the C++ source, but that goes beyond documentation :wink:.

bryanwweber:
@ischoegl Looking at the dev documentation here: https://cantera.org/documentation/dev/sphinx/html/cython/zerodim.html#massflowcontroller I still see `K_v` as the coefficient symbol. Do the Cython docs also need to be updated? I think people are much more likely to read the Python docs than C++, although I see that the symbols are consistent here with C++.

ischoegl:
@ bryanwweber ... Yes ... I just realized that I had looked at C++ as well (which is why I deleted my earlier declaration to be *done*). I agree that Python is the more likely avenue, and am mostly done with the updating the docs accordingly.

On a related note, what version of Sphinx are we on? Based on the [documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html), we may have access to `:py:method`, `:py:classmethod`, `:py:attribute`, etc.

ischoegl:
OK. I force-pushed - I don't think that `:py:var:` is the correct syntax for referencing class properties - please let me know.

Regarding the differences between C++ and Python, all coefficients are now defined for the parent object using the same `FlowDevice::m_coeff` (which is accessed using more appropriate names, e.g. 'Valve::setValveCoeff'), but for Python, they are set using properties, e.g. `Valve.valve_coeff`. There isn't an equivalent to a class property in C++, which explains the more generic formulation of the C++ docstrings.

ischoegl:
@bryanwweber ... finally located the reference I was wondering about: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

Updated accordingly, so I believe this should be done.

bryanwweber (approved):
Thanks @ischoegl!

bryanwweber:
FYI @speth this will require a few more changes to the docs from Cantera/cantera, I'm working on a PR. We should wait to merge this until after 2.5.1.

ischoegl:
ok. I am aware of remaining activity on the website. The only comment I have is that the new behavior will be default starting in 2.5.0, and I am not sure how long it would take before a minor release of 2.5.1.

bryanwweber:
@ischoegl thanks again for updating this. As you can probably see, we've tagged 2.5.1 already which incorporates the fixes to the docs. The problem wasn't with this code, but with the fact that Sphinx doesn't include inherited members by default, so the `*.set_...` (I forgot the name already 🤣) method from `FlowDevice` wasn't linking properly to the child classes. It's fixed in Cantera/cantera now, so this is all set.

ischoegl:
@bryanwweber .... Ok, thanks! I saw your fix earlier today. Just didn’t quite know a out the context earlier. All the best for the rollout!

PR #119, opened 2020-08-19
Blog post: GSoC 2020 Blog Post 4

paulblum:
Uploading my fourth blog post for my GSoC 2020 project, a walkthrough of my work so far with 0D steady-state `ReactorNet` solution.


posts/gsoc-2020-blog-4.md
```diff
@@ -0,0 +1,160 @@
...
+
+**ReactorNet.h**, line 24 (see this on [GitHub](https://github.com/paulblum/cantera/blob/ca36e253bd28c6d507eace5b6f1199cac64d8909/include/cantera/zeroD/ReactorNet.h#L24))
+
+```c++
+class ReactorNet : public FuncEval, public Cantera_NonLinSol
+```
+
+`ReactorNet` now exhibits *multiple inheritance*, since it was already a child of class `FuncEval`. The `FuncEval` parent class is used by `Integrator` during ODE time-integration, and its inheritance makes `ReactorNet` an *ODE right-hand-side evaluator*. (Confused? See my [last blog post](https://cantera.org/blog/gsoc-2020-blog-3)!)
+
+*Notice the difference in implementation here:* `Integrator` is an external solver object that uses user-defined functions from *another* class, `FuncEval`. This requires the inclusion of two headers, explicit initialization of an external object, and the passing of the `this` pointer:
```
decaluwe, line 40: @paulblum this is _really_ well done--great work!  One question, here, this paragraph here took me a while to parse, as the comparison you are making is not obvious.  When you say "notice the difference in implementation here," what is it you are comparing?  Your approach versus what?  Is there a way to make that clearer?

paulblum: Thanks @decaluwe! Just committed an updated version of the post, let me know if it's any clearer.

decaluwe: Looks great - thanks!

PR #124, opened 2021-01-15
Update requirements.txt and move to GitHub Actions

bryanwweber:
Still need to fix reference errors during the build. Have to debug locally though.


bryanwweber:
For some reason, GitHub Actions isn't running. Going to merge to `main` and see if that gets it to run.

PR #127, opened 2021-01-28
Updates for 2.5.0 release

### LINKED ISSUE #97: The minimum version of Python will be 3.5 from Cantera 2.5 and up
Cantera/cantera#693 requires Python 3.5 or higher. The compilation requirements currently list Python 3.3 as the minimum version:

https://github.com/Cantera/cantera-website/blob/d8b4f7b1bbcb615bbe701cf2ce20528524a94299/pages/compiling/installation-reqs.rst#L479

There may be other places this needs to be updated. This issue will catalog any others that are found on the website.

speth:
Just collecting some updates to be merged into the website in conjunction with the release of Cantera 2.5.0.

Some remaining To-do items:
- [x] The section "OS X & macOS" in installation-reqs.txt needs to be updated to reflect the end of Python 2 support, but I'm not actually sure what the right instructions here are.
- [x] Figure out what the minimum Numpy version is going to be for the Windows binaries

Resolves #97 
Resolves Cantera/cantera#565

### DIFF
--- a/pages/tutorials/MatlabTutorial.rst
+++ b/pages/tutorials/MatlabTutorial.rst
@@ -65,7 +65,7 @@ temperature, a pressure, species mole and mass fractions, etc. As we'll soon
 see, it has many more properties.
 
 The summary of the state of ``gas1`` printed above shows that new objects
-created from the ``gri30.cti`` input file start out with a temperature of 300 K,
+created from the ``gri30.yaml`` input file start out with a temperature of 300 K,
 a pressure of 1 atm, and have a composition that consists of only one species,
 in this case hydrogen. There is nothing special about H2, it just happens to
 be the first species listed in the input file defining GRI-Mech 3.0. In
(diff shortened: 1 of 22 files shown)


bryanwweber:
I'm working on the updates to the macOS section. Since we only require the Boost headers, what if we remove the recommendation for Homebrew? As best I can tell, that's the only thing that may be installed more easily from Homebrew. Python can come from python.org and all the other Python dependencies can come from `pip`.

If we suggest that people download the Boost headers and put them somewhere, we can eliminate installing Homebrew...

bryanwweber:
I pushed a commit that assumes we will keep Homebrew and just updates those instructions, pending my comment above.

decaluwe:
What about libomp?

bryanwweber:
> What about libomp?

Aww dangit. That and Git are probably easier to install from Homebrew. OK, I'm convinced, we'll keep it.

pages/tutorials/MatlabTutorial.rst
```diff
@@ -287,26 +287,15 @@ directory where they are located. Alternatively, you can call function
 
     >> adddir('/usr/local/cantera/my_data_files');
 
-XML files
-~~~~~~~~~
+Cantera currently supports three input formats for data. The primary format,
+introduced in Cantera 2.5.0, is the YAML format. Two older formats (CTI and XML)
+can also be used, but they are deprecated and support for them will be removed
+in a future release. Utilities are provided for converting existing CTI and XML
+files to YAML. For mor info, see
```
bryanwweber, line 294:
```suggestion
files to YAML. For more info, see
```

PR #132, opened 2021-02-17
Fix consecutive capitals in math mode

### LINKED ISSUE #128: "Caps" typogrify filter in Nikola is breaking rendering of LaTeX with strings of capital letters.
I notice in the [release notes of Nikola 8.1.2](https://getnikola.com/blog/#features-6) "Restore caps typogrify filter (wraps strings of capital letters with `<span class="caps">` (via Issue #3405)".

This is being applied to things like `RT` and `AB` which then breaks the mathjax rendering and screws up all the LaTeX math for any equations with two or more consecutive capital letters.
Screen Shot 2021-02-05 at 10 57 01 AM

Screenshot of documentation page https://cantera.org/science/reactions.html attached. OK math in green, broken math in red, with the offending letters circled. Same in Safari and Chrome.
Screen Shot 2021-02-05 at 10 52 05 AM

I don't know enough about Nikola to turn off this filter, but I think that's the problem.

bryanwweber:
Fixes #128. Pending resolution of getnikola/nikola#3512

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -698,8 +698,25 @@ OUTPUT_FOLDER = os.getenv("NIKOLA_OUTPUT_DIR", "output")
 # <https://getnikola.com/handbook.html#post-processing-filters>
 #
 from nikola import filters
+import typogrify.filters as typo
+from functools import partial
 
-FILTERS = {".html": [filters.typogrify, filters.add_header_permalinks]}
+FILTERS = {
+    ".html": [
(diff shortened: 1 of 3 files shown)


bryanwweber:
@12Chao I ended up having to really hack around this problem using `functools.partial`, which creates a copy of a function with some of the arguments "filled in".

PR #136, opened 2021-02-20
Render C++ and Fortran example programs

### LINKED ISSUE #73: Add C++ examples from Cantera/cantera
The C++ (and other language) examples from the Cantera/cantera/samples directory should be added to the website. The main question is how to show files that are split into header/source files? @speth Any thoughts?

speth:
Resolves #73.

~A first step towards resolving #73, although it chooses the "dumb" answer of just putting the header files on pages by themselves.~
The first version of this had headers and source files on separate files, but now they're combined.

~A companion update to `Cantera/cantera` ~should probably be introduced at some point to update the comments for each of these files, since the strategy used here of pulling the first comment in each example doesn't always provide a very good summary.~
Cantera/cantera#983 improves these docstrings to provide consistently useful titles and summaries.

~This is incomplete, because I haven't figured out how to get the C++ examples to show up as a category on the `examples/index.html` page. The `examples/cxx/index.html` page renders fine.~

### DIFF
--- a/plugins/render_cxx_examples.py
+++ b/plugins/render_cxx_examples.py
@@ -36,8 +36,8 @@ def render_example_index(site, kw, headers, output_file):
     """
     n = 3
     for head_dict in headers.values():
-        head_files = head_dict["files"]
-        head_dict["files"] = [
+        head_files = head_dict["names"]
+        head_dict["names"] = [
             head_files[i : i + n] for i in range(0, len(head_files), n)  # NOQA: E203
         ]
 
@@ -54,8 +54,8 @@ def render_example_index(site, kw, headers, output_file):
     site.render_template("cxx-example-index.tmpl", str(output_file), context)
 
 
-def render_example(site, kw, in_name, out_name):
-    """Render a single .m file to HTML with formatting.
+def render_example(site, kw, in_names, out_name):
+    """Render a set of .h and .cpp files to HTML with formatting.
 
     Parameters
     ==========
@@ -63,29 +63,32 @@ def render_example(site, kw, in_name, out_name):
         An instance of a Nikola site, available in any plugin as ``self.site``
     kw:
         A dictionary of keywords for this task
-    in_name:
-        The file to be rendered, as an instance of pathlib.Path
+    in_names:
+        The files to be rendered, as a list of pathlib.Path objects
     out_name:
         A pathlib.Path instance pointing to the rendered output file
 
     """
-    code = highlight(
-        in_name.read_bytes(), CppLexer(), utils.NikolaPygmentsHTML(in_name.name)
-    )
+    items = []
+    for source_file in in_names:
+        code = highlight(
+            source_file.read_bytes(),
+            CppLexer(),
+            utils.NikolaPygmentsHTML(source_file.name)
+        )
+        items.append({
+            "code": code,
+            "title": source_file.name,
+            "source_link": source_file.name,
+        })
 
-    title = in_name.name
-    permalink = out_name.relative_to(kw["output_folder"])
-    source_link = permalink.stem  # remove '.html'
     context = {
-        "code": code,
-        "title": title,
-        "permalink": str(permalink),
+        "items": items,
+        "title": out_name,
         "lang": kw["default_lang"],
-        "description": title,
-        "source_link": source_link,
         "pagekind": ["example"],
     }
-    site.render_template("examples.tmpl", str(out_name), context)
+    site.render_template("multifile-example.tmpl", str(out_name), context)
 
 
 class RenderCxxExamples(Task):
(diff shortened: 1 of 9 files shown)


plugins/render_fortran_examples.py
```diff
@@ -0,0 +1,252 @@
...
+    in_name:
+        The file to be rendered, as an instance of pathlib.Path
+    out_name:
+        A pathlib.Path instance pointing to the rendered output file
+
+    """
+    if str(in_name).endswith('.cpp'):
+        lexer = CppLexer()
+    else:
+        lexer = FortranLexer()
```
bryanwweber, line 75: The `FortranLexer()` appears not to like the format of our Fortran files... `c` lines aren't rendered as comments, and the `$` shows as an error. Is there a better lexer, or do we just have to fix up the examples?

plugins/render_fortran_examples.py
```diff
@@ -0,0 +1,252 @@
...
+                    in_block_comment = True
+                if in_block_comment or line.startswith('!') or line.startswith("//")  or line.startswith('* '):
+                    doc.append(line.lstrip('/* !'))
+                elif line.startswith('c     '):
+                    doc.append(line[6:])
+                elif doc:
+                    break
+            doc = ' '.join(doc).strip()
+
+            print(fortran_ex_file, '======\n', doc, '\n================\n')
```
bryanwweber, line 198: Extra `print()` here

plugins/render_fortran_examples.py
```diff
@@ -0,0 +1,252 @@
...
+            if not doc:
+                self.logger.warn(
+                    "The Fortran example {!s} doesn't have an appropriate summary. The "
+                    "first comment line of the Fortran file is taken as the "
+                    "summary.".format(fortran_ex_file)
+                )
+            name = fortran_ex_file.stem.replace("_", " ")
+            fortran_headers["examples"]["summaries"][fortran_ex_file.name] = doc
+            out_name = kw["output_folder"].joinpath(
+                self.examples_folder, fortran_ex_file.name + '.html'
```
bryanwweber, line 208: Same here about `Pathlib.with_suffix()`

bryanwweber (changes requested):
Thanks @speth, a long awaited feature! Is it possible to pair the header/source files and show them on the same page? I think it would need a new template, but that would be a little more natural than having separate pages for each of them.

speth:
Thanks for the review, @bryan. I was reminded of this issue because there's currently nowhere where Paul's new example shows up, unless you dig through the source code. 

Regarding the question of showing multiple examples on a page, I think part of the question there is how to know when to do that. For the C++ examples, I guess it's simple enough -- include all the files in each directory. 

But for the F77 sample, There are two separate Fortran programs that each rely on the `demo_ftnlib.cpp` file. Although as I think about it, I'm not sure how useful the `.cpp` file here is. I think you could maybe just call the low-level functions of the Cantera F90 interface and skip that anyway. And I'm not sure how concerned we need to be about highlighting support for F77. So for the Fortran examples, I'm kind of inclined to leave this with the "flat" structure for now, and hope that we ditch `demo_ftnlib.cpp` at some point.

bryanwweber:
Thanks @speth I agree that's a sensible approach!

plugins/render_fortran_examples.py
bryanwweber: Or do we just ignore it and move on? :joy:

plugins/render_fortran_examples.py
speth: I don't think that's what we want in this case - we're trying to generate a name like `isentropic.f.html` - we don't want to replace the existing `.f` suffix.

plugins/render_cxx_examples.py
```diff
@@ -0,0 +1,242 @@
...
+            if not doc:
+                self.logger.warn(
+                    "The C++ example {!s} doesn't have an appropriate summary. The "
+                    "first comment line of the C++ file is taken as the "
+                    "summary.".format(cxx_ex_file)
+                )
+            name = cxx_ex_file.stem.replace("_", " ")
+            cxx_headers["examples"]["summaries"][cxx_ex_file.name] = doc
+            out_name = kw["output_folder"].joinpath(
+                self.examples_folder, cxx_ex_file.name + '.html'
```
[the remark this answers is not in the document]
speth: After some of the other changes, this is now appending `.html` to what was the directory name, so I think using `with_suffix` is actually less clear.

speth:
Combining the C++ header and source file into a single example was more of a pain than I expected, but it's done now.

speth:
By th way, I think Cantera/cantera#983 should be merged first, so that when this builds and goes live, it will have the updated descriptions for these examples.

speth:
@bryanwweber I think this is ready, now that Cantera/cantera#983 has been merged.

bryanwweber (approved):
Thanks @speth!

PR #137, opened 2021-03-17
Update Gentoo Cantera package installation instructions page

band-a-prend:


pages/install/gentoo-install.rst
```diff
@@ -23,7 +23,7 @@ The following interfaces and tools are installed by default:
 
 - C++ Libraries and header files for compiling your own programs that use Cantera.
 
-- `CTI tools <../tutorials/ck2cti-tutorial.html>`__ (``cti`` USE flag, optional).
+- `YAML tools <../tutorials/ck2yaml-tutorial.html>`__ and legacy `CTI tools <../tutorials/ck2cti-tutorial.html>`__ (``cti`` USE flag, optional).
```
bryanwweber, line 26: Should this be changed to \`\`yaml\`\`?

bryanwweber (changes requested):
Just one question here, thanks for updating!

pages/install/gentoo-install.rst
band-a-prend: > (\``cti\`` USE flag, optional).

Do you mean this place?

Currently ``cti`` gentoo package option is used both for Gentoo Cantera 2.4.0 and 2.5.1 packages to provide tools for conversion Chemkin input files to Cantera Input format. 

Cantera 2.4.0 is still presented in gentoo repository at least until gentoo Cantera 2.5.1 stabilisation - last now is in testing branch. Although 2.4.0 doesn't provide mentioned `YAML tools`.

 For both versions the `cti` assumed as `Cantera Input` tools in general sense and just is specific repository convention.

bryanwweber: Yes, that's the place I meant. My understanding of CTI is "CanTera Input" _format_, hence why it was the file extension. I think it's probably fine to keep for now, but it might be good to add `yaml` as an alias, since the plan is to remove support for CTI format files eventually.

band-a-prend: > My understanding of CTI is "CanTera Input" format, hence why it was the file extension.

Yes, it's. While gentoo packaging the Cantera I required some short package flag to switch on/off installation of Cantera Input format converting tools that are provided by cantera SConstruct option `python_package="minimal"`. So it just a name of switcher in local gentoo build script. 

Don't worry about it - it's just inner workings. E.g. the option description [is here](
https://github.com/gentoo/gentoo/blob/master/sci-libs/cantera/metadata.xml#L16-L18) and the using of it [is here](https://github.com/gentoo/gentoo/blob/0c635a9098c3117fcc0695cd23b775ea56a4964d/sci-libs/cantera/cantera-2.4.0-r5.ebuild#L93-L98).


> I think it's probably fine to keep for now, but it might be good to add yaml as an alias, since the plan is to remove support for CTI format files eventually.

On Cantera 2.5 update I thinked about renaming this switcher to `ck2yaml` or even to `yaml` (the last  name of switcher is already  used by some other packages too so it will more suitable). But as Cantera 2.4 is stil presented in gentoo repository I awared of doing that during some transition period. Also I didn't want at that moment to create another one switcher with the same option. I'm planing to replace it later while dropping Cantera 2.4.0 from gentoo repository.

bryanwweber (approved):

pages/install/gentoo-install.rst
speth: Long term, I think it would be confusing to have the switch for whether or not to include various converters be named `yaml`, given that Cantera 2.5 now always supports mechanisms in the YAML format, whether or not you have any converters installed. Maybe `converters` would be a more appropriate name for this option in later versions? This is somewhat different from the CTI format, which does require conversion scripts to be present.

band-a-prend: @bryanwweber 
Thank you for merge!

@speth 
[There are many](https://www.gentoo.org/support/use-flags/) names for option switchers on gentoo. 
Usually `tools` is used in many packages as internal switch for conversion tools. Sometimes `utils` is used. There are global list of such switchers and local (i.e. individual for package). I  both case the description should be provided clarifiing the meaning of option.

Maybe `tools` or `utils` is more suitable for this case because their meaning are familiar for gentoo users assuming that some conversion tools or utils will be installed.

Yes, the `yaml` is mostly for YAML support option at all. But Cantera is always uses this format now as you mentioned. So in this situation yaml is percieved by me as new "Cantera Input" format just with differ structure and extension. I assume  because of that the `cti` option doesn't misslead me.

PR #157, opened 2021-10-20
Add Fedora Linux prebuilt packages

mefuller:
I have this working for the current development version of Cantera.
There's something in how paths are being written into the libraries in 2.5.1 as released that is causing those builds to fail.
This issue has been corrected in the current development version (2.6.0a3).


mefuller:
Cantera is now in the official Fedora repos and in EPEL for Enterprise Linux - what is the status of updating the installation instructions on the website (here or #164)?

PR #164, opened 2021-12-17
Update install instructions

### LINKED ISSUE #131: Add instructions for installing and using cantera-matlab conda package
Starting with Cantera 2.5, we now have Matlab packages which can be installed using Conda, (https://anaconda.org/Cantera/cantera-matlab/files) which work on Windows, macOS, and Linux. However, there are no instructions on how to install them, or how to use them once they are installed (i.e., what paths need to be set up in Matlab).

### LINKED ISSUE #154: Please add FreeBSD install instructions
Here: https://cantera.org/install/index.html

On FreeBSD Cantera is installed with the command: ```pkg install cantera```.

Thanks,
Yuri

bryanwweber:
Update the installation instructions for several new packages via Conda, FreeBSD, and Fedora/Enterprise Linux.

Closes #154 
Closes #131 
Merges #157


bryanwweber:
@yurivict @mefuller Can you please check this to make sure the instructions for your platforms are correct?

@decaluwe/@speth This is ready for review as part of the NSF annual report follow-up when you get a chance.

yurivict:
For FreeBSD: this statement is incorrect:
> A FreeBSD package is maintained by the Cantera community in the ``science`` port.

It should say
> A FreeBSD port science/cantera and a package are available.


> Note that the Matlab interface is not available from this port [...]

This option should be added instead of documenting that it is missing.
I tried to use octave, but it complains:
```
ERROR: Path set for 'matlab_path' is not correct.
ERROR: Path was: '/usr/local'
```
due to these lines:
```
    if not (os.path.isdir(matPath) and
            os.path.isdir(pjoin(matPath, 'extern'))):
        print("""ERROR: Path set for 'matlab_path' is not correct.""")
        print("""ERROR: Path was: '%s'""" % matPath)
        sys.exit(1)
```
Why does it expect the path to be ```*/extern``` ?

bryanwweber:
@yurivict thank you for the comments! I'll update the text accordingly.

> This option should be added instead of documenting that it is missing.\nI tried to use octave, 

If you'd like to support this I'm happy to update the docs. In the absence of a package, we should document that it is missing. Note that MATLAB and Octave are not the same, especially for compiled C code as used for the Cantera interface. If you'd like to add an Octave interface, that'd be an excellent feature!

yurivict:
> The MATLAB interface expects the MATLAB headers distributed by Mathworks, hence that check.

Is Octave missing support for some MATLAB features? What are these features? When I read about Octave vs. Matlab it said that they are functionally identical. I also know that another package, Dynare, links to Octave's headers fine.

bryanwweber:
@yurivict well, at least one difference is how the header files are distributed 🤪 a quick Google search says that `mex.h` is also distributed by Octave, so it may be possible to do a 1-to-1 substitution. Feel free to remove that check and see if things work!

yurivict:
See https://github.com/Cantera/cantera-website/issues/165

pages/install/fedora-rhel-install.rst
```diff
@@ -0,0 +1,58 @@
...
+
+      <h1 class="display-3">Installing on Fedora</h1>
+
+   .. class:: lead
+
+      RPM packages are provided for supported versions of Fedora Linux using a Community Projects
+      (COPR) repository. Note that the Matlab interface is not available from this archive;
+      to install the Matlab interface on Fedora, you must :ref:`compile the source code <sec-compiling>`.
+
+As of Cantera 2.6.0, packages are available for Fedora Linux 33, 34, 35, and
```
mefuller, line 19: Since I made the PR, F33 is no longer supported.
I'll update the language appropriately

pages/install/fedora-rhel-install.rst
```diff
@@ -0,0 +1,58 @@
...
+- ``cantera-devel`` - Libraries and header files for compiling your own C++ and
+  Fortran 90 programs that use Cantera.
+
+- ``cantera-common`` - Cantera data files and example programs
+
+To add the Cantera COPR:
+
+.. code-block:: bash
+
+   $ dnf copr enable fuller/Cantera
```
mefuller, line 40: There are two repos presently: one for Fedora and one for CentOS/EL
I'll also pass on an update

mefuller:
@bryanwweber I submitted a PR for the branch that you are proposing to merge here with a couple updates for the Fedora and EL instructions. I hope I'm doing this the right/efficient way

bryanwweber:
Thanks @mefuller that's perfect!

bryanwweber:
Thanks @yurivict and @mefuller I've updated the text with your suggestions. @Cantera/committers this is ready for review again.

mefuller (approved):
Looks great - thanks

pages/install/conda-install.rst
```diff
@@ -14,39 +14,58 @@
...
+
+Python interface
+================
+
+Cantera's Python interface is available from two channels:
+
+1. The official ``cantera`` channel. This channel should be used if you installed NumPy
+   from the default channel in conda. This channel also has pre-release versions of
+   Cantera for testing.
+2. The ``conda-forge`` channel. This channel should be used if you installed NumPy from
```
ischoegl, line 47: Not sure if I understand the difference in NumPy here. What combinations work? (It’s not clear if this is a preference or co-dependence).

pages/install/conda-install.rst
```diff
@@ -121,3 +140,22 @@ remove and then reinstall Cantera:
...
+
+.. code:: shell
+
+   conda create --name ct-env --channel cantera cantera cantera-matlab
+
+This will create an environment with Cantera's Python and MATLAB interfaces. Even if you
+plan to use Cantera via MATLAB, the Python interface is required to convert input files.
+In this case, Cantera must be installed from the ``cantera`` channel, so we add
+``--channel cantera`` and to tell Conda to look at the ``cantera`` channel in addition
+to the default channels.
```
ischoegl, line 161: Do people have to start MATLAB out of the conda environment? Are there any adjustments needed for matlabpath?

ischoegl (commented):
Thank you, @bryanwweber. This looks very good, but I have two minor comments.

pages/install/conda-install.rst
bryanwweber: The intention is that this is handled by a post-install script in the package which should be executed automatically. I'm not totally certain about whether or not post-install scripts are still executed by Conda or whether they've disabled that for security reasons.

pages/install/conda-install.rst
bryanwweber: Compilers and ABI compatibility. It actually applies to the base Python too, that might be a better way to state it, with reference to the Python rather than NumPy. Basically, there _might_ be compatibility problems with any compiled packages if you mix compilers. In practice I've never had any problems, but that's the gist of it. I guess it's really easier to say, use Conda forge for all the packages in an environment.

ischoegl: Ok, that's good to know. I guess what I was implying is that more direct instructions may be helpful in this write-up. Something along the lines: *In order to ensure compatibility of compiled code, it is recommended to install NumPy from `conda-forge`.*

pages/install/conda-install.rst
ischoegl: Thank you for the clarification. As the expected behavior isn't explicit (i.e. post-install scripts handle that everything is set up correctly), I was hoping that clarifying text could be added here.

pages/install/conda-install.rst
speth: The `conda-forge` channel also doesn't currently include Windows binaries, right?

bryanwweber: Yep, that's true too, thanks @speth

bryanwweber:
@speth / @ischoegl / @decaluwe Can you take a look at this so we can merge it?

ischoegl (approved):
This looks good to me.

PR #167, opened 2021-12-21
Ionflow

BangShiuh:
This PR describes the physics of class IonFlow which features the calculation of charged species distribution.

### DIFF
--- a/pages/science/flames.rst
+++ b/pages/science/flames.rst
@@ -96,7 +96,7 @@ differential equation for the scalar :math:`\Lambda`:
 Diffusive Fluxes
 ----------------
 
-This feature is only available when using class `IonFlow`. The species diffusive mass fluxes :math:`j_k` are computed according to either a
+The species diffusive mass fluxes :math:`j_k` are computed according to either a
 mixture-averaged or multicomponent formulation. If the mixture-averaged
 formulation is used, the calculation performed is:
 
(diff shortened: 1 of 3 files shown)


pages/science/flames.rst
```diff
@@ -125,6 +125,40 @@ where :math:`D_{ki}` is the multicomponent diffusion coefficient and
 :math:`D_k^T` is the Soret diffusion coefficient (used only if calculation of
 this term is specifically enabled).
 
+The Drift-Diffusion Model
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To account for the transport of charged species in a flame, the drift term is added to the
+diffusive fluxes of the mixture-average formulation according to [Ped1993]_,
```
speth, line 132: This is only done if the `IonFlow` domain is being used, right? I think that should be clarified.

pages/science/flames.rst
```diff
@@ -125,6 +125,40 @@ where :math:`D_{ki}` is the multicomponent diffusion coefficient and
...
+  
+   j_k^* = \rho \frac{W_k}{\overline{W}} D_{km}^\prime \frac{\partial X_k}{\partial z} +
+           s_k \mu_k E Y_k,
+
+where :math:`s_k` is the sign of charge (1,-1, and 0 respectively for positive, negative, and
+neutral charge), :math:`\mu_k` is the mobility, and :math:`E` is the electric field.
+The diffusion coefficients and mobilities of charged species can be more accurately calculated by
+**IonGasTransport::getMixDiffCoeffs** and **IonGasTransport::getMobilities**. In addition, the
+correction can only be applied to the fluxes of neutral species to ensure the accuracy of the
+fluxed of charged species,
```
speth, line 144: The wording of this is a bit confusing, and I have a couple questions about this. First, I'd try something more like
> In addition, the following correction is applied to the fluxes of neutral species only, to preserve the correct fluxes of charged species:
But beyond that, I'm wondering, does this correction still preserve the condition that the sum of the diffusive mass fluxes is zero?

pages/science/flames.rst
```diff
@@ -125,6 +125,40 @@ where :math:`D_{ki}` is the multicomponent diffusion coefficient and
...
+
+.. math::
+
+    j_k = j_k^* - \frac {1 - |s_k|} {1 - \sum_i |s_i| Y_i} Y_k \sum_i j_i^*.
+
+The equation for electrostatics is added to obtain the electric field,
```
speth, line 150: The equation is added _where_?

Currently, this is mixed in with the discussion of diffusion terms. If you want to fully integrate the description of ionized flows with the "normal" flame model, I think this would need to be introduced earlier, alongside the species / energy equations.

The other option would be to have a specific section after everything about normal flames (that is, after all the description of all the boundary conditions) to introduce the additional considerations for ionized flames. I don't have a strong preference one way or the other.

pages/science/flames.rst
```diff
@@ -125,6 +125,40 @@ where :math:`D_{ki}` is the multicomponent diffusion coefficient and
...
+
+.. math::
+
+    \frac{\partial E}{\partial z} = \frac{e}{\epsilon_0}(n_+ - n_-),
+
+    E|_{z=0} = 0,
```
speth, line 156: This isn't just the number density of each charged species, right? It would depend on the charge of the specific species if you had an ion with charge +2, right? Can you write this more explicitly in terms of the mass/mole fractions and species charges?

speth (changes requested):
Just had a few suggestions. If you want to update this PR to include documentation related to the `TwoTempPlasmaReaction` work that was just merged into the `main` branch, that would be helpful as well.

pages/science/flames.rst
BangShiuh: Yes, it only corrects the flux of neutral species to ensure zero diffusive flux.

pages/science/flames.rst
BangShiuh: ok

speth (approved):
Thanks, @BangShiuh, this looks good. Sorry for the delay in getting this merged.

PR #171, opened 2022-01-17
Update Jupyter examples

speth:
- Improve descriptions given in the index of Jupyter examples (in combination with Cantera/cantera-jupyter#33). For example, this:
image
becomes:
image

- Automatically link to any input files used in each example that come from the `data` directory (also reorganized by Cantera/cantera-jupyter#33). For example:
image

### DIFF
--- a/themes/cantera/templates/jupyter-example-index.tmpl
+++ b/themes/cantera/templates/jupyter-example-index.tmpl
@@ -15,7 +15,7 @@ just want to experiment, you can give Cantera a test drive in the cloud. Click o
 below to launch an interactive environment where you can run these examples. For this, there is no
 installation required, but you cannot save and resume your work.</p>
 
-<a href="https://mybinder.org:/repo/cantera/cantera-jupyter" rel="nofollow">
+<a href="https://mybinder.org/v2/gh/Cantera/cantera-jupyter/main" rel="nofollow">
   <img src="https://camo.githubusercontent.com/70c5b4d050d4019f4f20b170d75679a9316ac5e5/687474703a2f2f6d7962696e6465722e6f72672f62616467652e737667" alt="Binder" data-canonical-src="https://mybinder.org/badge.svg" style="max-width:100%;">
 </a>
 
--- a/themes/cantera/templates/jupyter-example-index.tmpl
+++ b/themes/cantera/templates/jupyter-example-index.tmpl
@@ -12,7 +12,10 @@ the "Source" link at the top of each example page and you should be able to run 
 some examples, you will also need to download the additional data files listed below the
 "Source" link. These data files should either be placed in a subdirectory named "data"
 one level up from the directory containing the notebook file, or the notebook can be
-modified to point to the location of these data files on your computer.
+modified to point to the location of these data files on your computer. Alternatively,
+you can download all of these examples and their accompanying data files by cloning the
+<a href="https://github.com/Cantera/cantera-jupyter">cantera-jupyter</a> repository from
+GitHub.
 </p>
 
 <p><strong>New Cantera Users</strong>: If you don't have an existing Cantera installation and you
--- a/themes/cantera/templates/jupyter-example-index.tmpl
+++ b/themes/cantera/templates/jupyter-example-index.tmpl
@@ -31,7 +31,7 @@ installation required, but you cannot save and resume your work.</p>
       {% set fname = file.name %}
       <!-- <div class="col-sm-4"> -->
         <div class="card">
-          <a href="{{ head }}/{{ fname }}.html"><h5 class="card-header">{{ fname }}</h5></a>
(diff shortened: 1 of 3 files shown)


ischoegl (changes requested):
@speth ... thank you for this update. Just some minor comments from my side ...

I just pulled the artifacts and the main page looks great! Only thing I noticed is that one of the newer examples isn't added yet: [ck2yaml_demo](https://github.com/Cantera/cantera-jupyter/blob/main/input/ck2yaml_demo.ipynb)

Beyond, I think that locations of YAML files could also be pointed out [here](https://cantera.org/examples/jupyter/index.html), i.e.

> If you have Cantera and the Jupyter Notebook server installed on your local machine, simply download any Jupyter notebook by clicking the "Source" link at the top of each example page and you should be able to run it.

[As an aside, if I follow these instructions on Firefox/Fedora, I get to some cryptic looking output for the artifact when I click on 'source', although I didn't experience this on the actual website.]

Spot-checking, the data files are displayed correctly for `lithium_ion_battery.ipynb`, but are missing for `equations_of_state.ipynb`. On that note, the latter notebook also has some problems with LaTeX rendering that I didn't notice elsewhere.

On `interactive_path_diagram.ipynb`, the rendering of some of the headers is not pretty (likely the wrong header level). 

Some of the issues I am pointing out are obviously not website related, but should be fixed by a minor update on `cantera-jupyter`.

bryanwweber:
> Some of the issues I am pointing out are obviously not necessarily website related, but should be fixed by a minor update on `cantera-jupyter`.

I'm happy to add these to https://github.com/Cantera/cantera-jupyter/pull/34

speth:
I fixed the missing `ck2yaml` example (a new category should have been added when that example was added), added directions explaining that the data files also need to be downloaded, and fixed the cases where the data files weren't being detected.

One case that I did _not_ resolve is the input files for the `ck2yaml` example, which breaks the pattern of using the `data` directory for all input files. I think those input files should be moved to conform to the pattern of always using the `data` directory, but that needs to be dealt with on the `cantera-jupyter` repository.

I also did not see a rendering problem with `equations_of_state.ipynb` - could you be more specific (not that I think that has anything to do with this PR).

ischoegl:
> I fixed the missing `ck2yaml` example (a new category should have been added when that example was added) [...]

Thank you! Example was merged very recently (after creation of this PR!), although I wrote it quite some time ago. Honestly didn't think about it until it came up here ...

> One case that I did not resolve is the input files for the `ck2yaml` example, which breaks the pattern of using the data directory for all input files. I think those input files should be moved to conform to the pattern of always using the data directory, but that needs to be dealt with on the `cantera-jupyter` repository.

I think that fell between the cracks - changes date back way before Cantera/cantera-jupyter#33, although both PR's were merged the same day. I am hesitant to fix this now as it will mess with Cantera/cantera-jupyter#34. Imho, there should be some distinction as CK input is a separate category though.

> I also did not see a rendering problem with `equations_of_state.ipynb` - could you be more specific (not that I think that has anything to do with this PR).

I checked again today, and it may have something to do with my local OS. When viewing the downloaded artifacts yesterday, I had stray `$`'s in my output, but they're gone today.

ischoegl (approved):

themes/cantera/templates/jupyter-example-index.tmpl
```diff
@@ -6,16 +6,21 @@
...
-<p><strong>Existing Cantera users</strong>: If you have Cantera and the Jupyter Notebook server
-installed on your local machine, simply download any Jupyter notebook by clicking the "Source" link
-at the top of each example page and you should be able to run it.</p>
+<p><strong>Existing Cantera users</strong>: If you have Cantera and the Jupyter Notebook
+server installed on your local machine, simply download any Jupyter notebook by clicking
+the "Source" link at the top of each example page and you should be able to run it. For
+some examples, you will also need to download the additional data files listed below the
+"Source" link. These data files should either be placed in a subdirectory named "data"
+one level up from the directory containing the notebook file, or the notebook can be
+modified to point to the location of these data files on your computer.
```
ischoegl, line 15: On second thought: would it make sense to just tell users to clone the `cantera-jupyter` repo?

speth: No, I don't think we want to put the barrier of using Git between our users and some of our most readily accessible examples.

ischoegl: Ok. In this case I’m ok with merging … :shipit:

speth: Well, on second thought, I don't think it hurts to mention the option of cloning the repo as an alternative.

bryanwweber:
> I fixed the missing `ck2yaml` example (a new category should have been added when that example was added),

Sorry, I should have caught that at review.

> but that needs to be dealt with on the `cantera-jupyter` repository.

Will do.

bryanwweber (approved):

PR #174, opened 2022-02-11
New section for Equation of state documentation in the Science section

gkogekar:
Created a new page under the Science section on Ideal and non-Ideal equations of state, with some references.  The page contains a very basic information on how thermodynamic properties are calculated. The detailed versions of mathematical derivations for Redlich--Kwong and Peng--Robinson need to be uploaded somewhere on the website.


speth:
Thanks for starting in on this, @gkogekar. I had couple of thoughts on the high-level organization of this material.

Instead of adding a new top-level category within the "Science" section, I think this information all belongs in the "[Phases](https://cantera.org/science/phases.html)" section, since the goal is to provide documentation for certain `ThermoPhase` implementations. What I was imagining is that the "Phases" page would mainly be another index-like page, with cards linking to documentation for particular phase models. It might make sense to group closely-related models like Redlich-Kwong and Peng-Robinson under a single page, though.

I would say that basically everything that's currently on the "Phases" page would be better off somewhere else. I'd be curious if others (paging @decaluwe, @bryanwweber, @ischoegl) have thoughts on the organization as well.

As a minor point, could you rebase your branch on the current main branch rather than generating an unnecessary merge commit?

ischoegl:
@gkogekar ... likewise, thanks for starting here! I overall agree with @speth in creating another index-like page. 

Fwiw, I am linking  Cantera/enhancements#6 (specifically step 3), although this goes well beyond of what would be expected from a single PR :joy:

decaluwe:
Thanks, @ischoegl for linking.  I need to clear a day or two and finally finish this task. 

While I agree that issues goes beyond the scope of what @gkogekar is doing here, I do think it is highly relevant - whatever page structure we set up in this PR should have in mind what we want to tackle in https://github.com/Cantera/enhancements/issues/6 so that those additions can flow smoothly without undoing anything we do here.

speth:
Superseded by https://github.com/Cantera/cantera/pull/1647

PR #178, opened 2022-03-13
Installation-related updates

### LINKED ISSUE #142: Installation instructions should explain how to upgrade
At least for the Conda and Ubuntu packages, the installation instructions should provide the commands to upgrade to the latest version of Cantera.

### LINKED ISSUE #143: Conda installation instructions should mention 64-bit requirement
As long as Anaconda and Miniconda are distributing 32-bit installers, we should clearly indicate that you have to use a 64-bit conda to install Cantera.

### LINKED ISSUE #166: Differentiate installation instructions by API
At the moment, [installation instructions](https://cantera.org/install/index.html) are not specific to the front end used by a user. Especially for new users, it may make sense to provide instructions that differentiate by API, i.e. currently

* Python
* Matlab
* C++/Fortran

Per comment by @speth, the 'best' option depends on the API, where for Python, users are likely best served by using `conda` (or the newly added `pip`), whereas system installers or compilation from source may make most sense for other cases:

> maybe "How do you want to use Cantera" needs to be the first level distinction on the "Installing Cantera" page: If your answer is "I want to use Cantera solely from Python", then being pointed to pip/Conda first makes a lot of sense.

See also: #164

speth:
- Fixes #142
- Fixes #143 
- Fixes #166


pages/install/install.rst
```diff
@@ -14,149 +14,55 @@
...
+Installing the Cantera Python Interface
+=======================================
 
-   .. row::
+- The easiest way to install the Cantera Python interface on all operating systems is by
+  using :ref:`Conda <sec-install-conda>`.
 
-      .. container:: col-12
+- Windows users can use
+  :ref:`MSI installer packages <sec-install-windows>`.
```
bryanwweber, line 28: In Cantera/Cantera#1154 the MSI installers are replaced with wheels. Assuming that we intend to merge that prior to 2.6, this will need to be updated.

speth: My aim with this was to focus on the organization of these docs and write something that was correct now, not to provide all of the necessary updates for the 2.6 release, which will involve a lot of other changes.

I don't actually fully know what the plan is for the wheel-based installation -- are we just going to suggest  using `pip`, or do we want to provide instructions for installing wheels directly?

(I think you meant Cantera/cantera#1158, right?)

bryanwweber: > (I think you meant [Cantera/cantera#1158](https://github.com/Cantera/cantera/pull/1158), right?)

Oops, yes, thank you. Fair enough about limiting the changes.

As far as I know, the instructions for installing wheels directly is the same as installing from `pip`, except that you specify the wheel filename you've downloaded instead of just `cantera`.

speth:
I'd like some feedback on this reorganization of the installation instructions, and to resolve this PR before moving ahead with updating the installation docs for the 2.6.0 release.

pages/install/conda-install.rst
```diff
@@ -16,24 +16,32 @@
...
-Anaconda and Miniconda both include ``conda``; the difference is that Anaconda includes a large
-number of Python packages that are widely used in scientific applications, while Miniconda is a
-minimal distribution that only includes Python and Conda, although all of the packages available in
-Anaconda can be installed in Miniconda. For more details on how to use conda, see the `conda
+      install Cantera.
+
+Both the Anaconda and Miniconda distributions are available for Linux, macOS (Intel and
+ARM/M1), and Windows. Note that installing Cantera using Conda will only provide
+the Cantera :ref:`Python <sec-conda-python-interface>` or :ref:`MATLAB
+<sec-conda-matlab-interface>` interfaces. If you want to use the other Cantera
```
ischoegl, line 24: If we can get `libcantera-devel` to properly compile on `conda` then we have headers and libraries also. These work well on Linux, and I also have some initial results on Windows.

pages/install/conda-install.rst
```diff
@@ -16,24 +16,32 @@
...
+interfaces (to use Cantera from Fortran, C++, or C) then see the :ref:`OS-specific
+installation options <sec-install>`. Windows users should install a 64-bit version of
+Anaconda or Miniconda, since the Cantera Conda packages are only available for 64-bit
+installations.
+
+Both Anaconda and Minconda include the ``conda`` package manager; the difference is that
```
ischoegl, line 30:
```suggestion
Both Anaconda and Miniconda include the ``conda`` package manager; the difference is that
```

ischoegl (commented):
I only have minor comments.

pages/install/install.rst
```diff
@@ -14,149 +14,55 @@
 
    .. class:: lead
 
-      The following instructions detail how to install Cantera on a variety of platforms. We highly
-      recommend that all new users install the Python interface via Conda. If you want to use one of
-      the other interfaces, please select the installation instructions for your platform.
+      The following instructions detail how to install the different Cantera interfaces
+      on a variety of platforms. We highly recommend that all new users install the
+      Python interface via :ref:`Conda <sec-install-conda>`.
```
decaluwe, line 19: Since users can also do the Matlab interface from Conda, now, would something like

> We highly recommend that all new users seeking a Python or Matlab interface install Cantera via :ref:`Conda <sec-install-conda>`.

work?

pages/install/conda-install.rst
speth: I think there's more to do as far as documenting `libcantera-devel`, so I've created a separate issue for that (#180).

pages/install/install.rst
speth: Given the many shortcomings of the current Matlab toolbox, I would not recommend it on an equal footing with the Python module. This isn't just a question of what's easiest to install.

pages/install/conda-install.rst
ischoegl: Sounds good. That's all i Have then!

ischoegl (approved):
Looks good to me …

speth:
Thanks, but CI is now failing somewhere inside of Nikola, which shouldn't have anything to do with this PR.

ischoegl:
> Thanks, but CI is now failing somewhere inside of Nikola, which shouldn't have anything to do with this PR.

… had everything typed up right when you pushed …

speth:
OK, it's all working again!

PR #179, opened 2022-03-22
WIP: Reorganize Science section.

decaluwe:
Related to https://github.com/Cantera/enhancements/issues/6 and https://github.com/Cantera/enhancements/issues/68.

Change the structure of the `Science` section so we can more easily add important details.

Putting this general structure out here now for comment.  If people are on board, it will be relatively straightforward to first re-populate this with what the `Science` section already has. After that, we will have a more intuitive place and structure for adding more details.


ischoegl:
Thanks, @decaluwe! Two thoughts:

1. On the upside, having the science section exactly map `Thermo` / `Kinetics` / `Transport` would be nice as it neatly maps the C++ core implementation (as well as other API's).
2. On the flipside, it isn't entirely clear to me that differentiating thermo-aspects and transport aspects of `Species` is necessarily intuitive.

From a user perspective, the differentiation in (1) is not obvious as this really only happens 'under the hood'. The thought-process in (2) circles back to what is currently already in place. 

Ultimately, a user mostly sees `Solution`s, which are built from `Species` and `Reaction`, so it makes sense to feature them at the top level. Personally, I view `Solution` mostly equivalent to `Phase` (if it weren't for `PureFluid`, which imho really should be `Mixture`). So I'd vote to stick with the current organization. Just my 2 cents of course!

bryanwweber (changes requested):
Thanks @decaluwe! I like this reformatting to be around the scientific topics rather than the implementation details. I agree with @ischoegl that we need to include the relationship to the implementation _somewhere_ although I'm fine with deferring the decision about where that should end up exactly. The line comments below are hopefully more general thoughts (other than the few typos) that I hope can be applied going forward.

decaluwe:
Thanks, @ischoegl and @bryanwweber!  @ischoegl I agree that there will be certain redundancies, in talking about species - phase interactions between transport and thermo.  But I think that burden falls on the author, while not really burdening a reader.  My thought for this is sort of like a companion to the API documentation, but for the theory.  Somebody consulting this will be seeking out a specific bit of theory for a calculation they are doing.  Organizing it by thermo / kinetics / transport, _I think_, gives them an intuitive way to downselect/filter.  

I'd actually like to make what is currently there a little _less_ like the content at https://cantera.org/tutorials/input-files.html, and make this a stopping point purely for those wanting to know the theory.  Some of the stuff at the top of the "species" page, in particular, doesn't really qualify as "theory," to me...

@bryanwweber thanks for the input.  Some of the comments about wanting more specific info I'm a little ambivalent on - the intent is to have the users click on the cards for the actual explanation, but just wanted to give them an overview, here.  But I'll take a crack at it and see if I can address your concerns.  Regarding P-v-T, I can add more detail, but to me an EoS does two things that are somewhat loosely related: (i) an overall governing equation for how P, v (or rho), and T are related, which most would actually call the equation of state, and (ii) a set of relationships for more "thermo-y" properties (u, h, s, mu, etc.)

Anyway, I'll put some more work into this today and see where I get.

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,60 @@
...
+   .. raw:: html
+
+      <h1 class="display-3">Calculating thermodynamic properties in Cantera</h1>
+
+   .. class:: lead
+
+      Here, we describe how Cantera uses species and phase information to calculate thermodynamic properties. Thermodynamic properties typically depend on information at both the species and phase levels.
+
+      - The user must specify a thermodynamic model for each species and provide inputs that inform how species-specific properties are calculated (e.g. as a function of temperature).
+      - The user also selects a phase model. This model describes how the species interact with one another to determine overall phase properties. This includes general :math:`P-v-T` behavior, as well as how species-specific properties are used to calculate phase-average properties such as internal energy, entropy, etc.
```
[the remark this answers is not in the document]
decaluwe: My intent is really as simple as "for example, calculate the phase pressure for a given composition, temperature, and total molar volume." 

I can probably add it as a parenthetical.

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,60 @@
...
+
+   .. raw:: html
+
+      <h1 class="display-3">Calculating thermodynamic properties in Cantera</h1>
+
+   .. class:: lead
+
+      Here, we describe how Cantera uses species and phase information to calculate thermodynamic properties. Thermodynamic properties typically depend on information at both the species and phase levels.
+
+      - The user must specify a thermodynamic model for each species and provide inputs that inform how species-specific properties are calculated (e.g. as a function of temperature).
```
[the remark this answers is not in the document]
decaluwe: > I prefer to avoid Latin abbreviations

👍🏼

> The other thing on this line is it might be nice to clarify which are species-specific properties, since you've clarified which ones are phase-average properties.

I can add some examples (reference Cp, h, and s values as a function of T should cover it).

Regarding giving a sense of how species and phase properties relate: I feel like there's a catch-22, here: anything that I can describe as being _generally_ true here would not be something that a phase model provides (if it is always true, then it can just be implemented generally for all `Phase` models).

decaluwe:
Lol, ended up adding much more detail than I thought I would.  @ischoegl and @bryanwweber, LMK what you think.  If 👍 I'll move forward with repopulating these sections with existing content.

bryanwweber (commented):
Thanks @decaluwe! I like where this is heading, looking forward to seeing more 😄

pages/science/index.rst
```diff
@@ -37,56 +36,55 @@ technology applications.
...
+               The theory behind how Cantera calculates species and phase thermodynamic properties.
 
       .. container:: card
 
          .. container::
             :tagname: a
-            :attributes: href=science-species.html
-                         title=Species
+            :attributes: href=reactions.html
+                         title=Reactions
```
ischoegl, line 58:
```suggestion
            :attributes: href=kinetics.html
                         title=Kinetics
```

ischoegl (commented):
Friendly amendment: I believe `Kinetics` may be a better top category than `Reactions`, as some of the kinetics handling goes beyond individual reactions.

pages/science/index.rst
speth: Could I suggest "Kinetics and Reaction Rates"? That keeps the breadth, but also hints at the fact that much of the description to follow will be about the different reaction rate parameterizations available.

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,89 @@
...
+
+   - The user must specify a thermodynamic model for each species and provide inputs that inform how species-specific properties are calculated. For example, the user specifies how the reference enthalpy and entropy values for each species are calcualted, as a function of temperature.
+   - The user also selects a phase model. This model describes how the species interact with one another to determine phase properties and species specific properties, for a given thermodynamic state. This includes general :math:`P-v-T` behavior (for example, calculate the phase pressure for a given molar volume, temperature, and chemical composition), as well as how species-specific properties, such as internal energy, entropy, and others depend on the state variables
+
+Example: The Ideal Gas Model
+============================
+For a simple example: in the Ideal Gas model, one might use 7-parameter NASA polynomials to specify the species thermodynamics.  These would be used to calculate the reference molar enthalpy :math:`h_k^\circ(T)` and entropy :math:`s_k^\circ(T)` for a given species :math:`k` as a function of temperature :math:`T`:
```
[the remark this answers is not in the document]
bryanwweber: > As far as Cantera is concerned, I don't think there is such a thing as an "ideal gas species model", though

Hmmm, fair enough. Maybe we should choose a different example here, which may work if we limit the equations? Perhaps R-K or P-R?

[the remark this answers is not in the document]
decaluwe: @bryanwweber -- re: using R-K or P-R.  

a. Those would also rely on NASA polynomials, most frequently.

b. My intent here was for simplicity.  Show that the `species` model is used to find the "reference state" properties, and that the `phase` model uses this reference properties to then find the actual properties. I feel like Ideal Gas is both the most familiar and computationally "cleanest" way to demonstrate this.

However, it seems like you might be after something else - can you clarify?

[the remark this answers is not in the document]
decaluwe: I'm fine going with `\hat`.  Not sure if I've got the chops to configure Mathjax, but I'll give it a look.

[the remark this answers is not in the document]
bryanwweber: > a. Those would also rely on NASA polynomials, most frequently.

Dang.

> b. My intent here was for simplicity. Show that the species model is used to find the \"reference state\" properties, and that the phase model uses this reference properties to then find the actual properties.

That's actually a really good statement of what I think I'm after!

> I feel like Ideal Gas is both the most familiar and computationally \"cleanest\" way to demonstrate this.

Agreed. 

> However, it seems like you might be after something else - can you clarify?

I'm not totally sure I know, except that I'll know it when I see it 🤪 more seriously, I think this is very close to what I'm thinking of. I think it's tilted a little too far into details in this iteration. My comment was an inaccurate attempt to simplify, and also to connect to the prior point that phase and species thermo models have to be compatible. That last bit, the connection, is probably just asking too much of the text here though, so I think we can drop that attempt. But I do think it needs to be simplified in the direction my comment was trending.

I'll take a look at the configuration for Mathjax when I'm back at my computer in the morning my time 😊 it should be similar to defining a macro in LaTeX, mostly a matter of getting it into the right spot.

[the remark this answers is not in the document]
decaluwe: Hmm, so does that mean that the molar volume is `\hat{v}`?  That feels... odd, to me, but it technically the consistent approach.

pages/science/index.rst
decaluwe: I like titling the card "Kinetics and Reactions Rates," but switching the html tag to the shorter "kinetics" feels like the right move, to me...

decaluwe:
Okay, I think this stage of what I had in mind is complete, for now.
1. Please check my ideal gas example.  I originally had `u_k = u^o_k - PvX_k`, but I think the X_k is incorrect/unnecessary (as a check: u for the phase should be recovered from sum_k(u_k * X_k)).
2. My next step is to move the `Transport` content out of `Thermodynamics/species` and into the (currently blank) `Transport` page. My thinking is that the landing page for Transport will be very similar to what I currently have for Thermo.

decaluwe:
@bryanwweber I think this is ready for a review.  Mainly, I don't know if the thermo page hits what you had in mind.

As noted in my last commit message, the Transport landing page is a bit underbaked.  But as it fully reproduces all the content on the current website, I'd like to stop here, get this PR merged, and then move forward on a fresh PR once the overall structure is approved.  This

decaluwe:
I'll resolve the conflict in `reactions.rst` tomorrow...

bryanwweber:
@decaluwe When you've resolved the conflict can you request a review from me? Thanks :smile:

decaluwe:
Hmm, kind of stumped, here, as `reactions.rst` no longer exists.  I've reverted my last commit, locally, to see if the conflict still exists, but can't seem to push the reverted branch to my repo (I just get `Everything up-to-date`).  

Any advice/tips/tricks?

speth:
I just rebased this onto the current `main` branch and pushed it to my fork. You can check that that still has all of your intended changes and then either fetch my branch and do a hard reset of your branch and then force push here, or if you want I can do a force push here and then you can fetch and do the local hard reset.

decaluwe:
Thanks, @speth -- worked like a charm.

Okay, @bryanwweber -- have at it 😉

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,81 @@
...
+      
+   Thermodynamic properties typically depend on information at both the species and phase levels. The user must specify thermodynamic models for both levels, and these selections must be compatible with one another. For instance: one cannot pair certain non-ideal species thermodyamic models with an ideal phase model.
+
+   - The user must specify a thermodynamic model for each species and provide inputs that inform how species properties are calculated. For example, the user specifies how the reference enthalpy and entropy values for each species are calcualted, as a function of temperature.
+   - The user also selects a phase model. This model describes how the species interact with one another to determine phase properties and species specific properties, for a given thermodynamic state. This includes general :math:`P-\hat{v}-T` behavior (for example, calculate the phase pressure for a given molar volume, temperature, and chemical composition), as well as how species-specific properties, such as internal energy, entropy, and others depend on the state variables
```
bryanwweber, line 18:
```suggestion
   - The user also selects a phase model. This model describes how the species interact with one another to determine phase properties and species specific properties, for a given thermodynamic state. This includes general :math:`P`-:math:`\hat{v}`-:math:`T` behavior (for example, calculate the phase pressure for a given molar volume, temperature, and chemical composition), as well as how species-specific properties, such as internal energy, entropy, and others depend on the state variables
```

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Example: The Ideal Gas Model
+============================
+For a simple example: in the Ideal Gas model, one might use 7-parameter NASA polynomials to specify the species reference thermodynamic quantities.  These would be used to calculate the reference molar enthalpy :math:`\hat{h}_k^\circ(T)` and entropy :math:`\hat{s}_k^\circ(T)` for a given species :math:`k` as a function of temperature :math:`T`. See the `NASA Polynomials Species Thermo entry </science/science-species.html#the-nasa-7-coefficient-polynomial-parameterization>`__ for more information.
+
+At the phase level, the Ideal Gas Law provides the P-v-T relationship, called an equation of state. This is used, for example, to calculate the pressure as a function of molar volume :math:`\hat{v}` and temperature :math:`T`:
```
bryanwweber, line 24:
```suggestion
At the phase level, the Ideal Gas Law provides the :math:`P`-:math:`\hat{v}`-:math:`T` relationship. The ideal gas law is an example of an equation of state. This is used, for example, to calculate the pressure as a function of molar volume :math:`\hat{v}`, and temperature, :math:`T`:
```

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. math::
+   p = \frac{\overline{R}T}{\hat{v}}
+
+where :math:`\overline{R}` is the Universal Gas Constant. Maxwell's relations are used to derive other thermodynamic properties from the equation of state. With the Ideal Gas phase model, these reduce to rather simple forms. For example, for a species :math:`k`, the Ideal Gas molar internal energy :math:`\hat{u}_k` and entropy :math:`\hat{s}_k` are:
```
bryanwweber, line 29:
```suggestion
where :math:`\overline{R}` is the Universal Gas Constant. The `Maxwell relations <https://en.wikipedia.org/wiki/Maxwell_relations>`__ are used to derive other thermodynamic properties from the equation of state. With the Ideal Gas phase model, these reduce to rather simple forms. For example, for a species :math:`k`, the Ideal Gas molar internal energy :math:`\hat{u}_k` and entropy :math:`\hat{s}_k` are:
```

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,81 @@
...
+            .. container:: card-header section-card
+               :tagname: div
+
+               Species
+
+         .. container:: card-body
+
+            .. container:: card-text
+
+               The models and equations that Cantera uses to calculate species thermodynamic properties.
```
bryanwweber, line 61:
```suggestion
               The models and equations that Cantera uses to calculate species thermodynamic properties, such as the NASA 7-parameter polynomial form.
```

pages/science/thermodynamics.rst
```diff
@@ -0,0 +1,81 @@
...
+            .. container:: card-header section-card
+               :tagname: div
+
+               Phases
+
+         .. container:: card-body
+
+            .. container:: card-text
+
+               The theory behind some of Cantera's phase models.
```
bryanwweber, line 79:
```suggestion
               The theory behind some of Cantera's phase models, such as the Ideal Gas Law.
```

pages/science/transport.rst
```diff
@@ -0,0 +1,49 @@
...
+
+   .. raw:: html
+
+      <h1 class="display-3">Calculating transport properties and rates in Cantera</h1>
+
+   .. class:: lead
+
+      Here, we describe how Cantera uses species and phase information to calculate transport properties and rates.
+
+   Similar to Cantera's approach to `thermodynamic properties </science/thermodynamics.html>`__, transport property calcualtions in Cantera depend on information at both the species and phase levels. The user must specify transport models for both levels, and these selections must be compatible with one another.
```
bryanwweber, line 15:
```suggestion
   Similar to Cantera's approach to `thermodynamic properties </science/thermodynamics.html>`__, transport property calculations in Cantera depend on information at both the species and phase levels. The user must specify transport models for both levels, and these selections must be compatible with one another. 
```

pages/science/species-thermo.rst
```diff
@@ -80,25 +80,14 @@ expressions to compute the properties, they all require thermodynamic property
 information for the individual species. For the phase types implemented at
 present, the properties needed are:
 
-1. the molar heat capacity at constant pressure :math:`\hat{c}^0_p(T)` for a
```
bryanwweber, line 83: There's a link a few lines up from here which is broken now, it should be `/science/phase-thermo.html` I think.

pages/science/species-thermo.rst
```diff
@@ -80,25 +80,14 @@ expressions to compute the properties, they all require thermodynamic property
...
-2. the molar enthalpy :math:`\hat{h}(T_0, P_0)` at :math:`P_0` and a reference
-   temperature :math:`T_0`;
-3. the absolute molar entropy :math:`\hat{s}(T_0, P_0)` at :math:`(T_0, P_0)`.
+1. the molar heat capacity at constant pressure :math:`\hat{c}^\circ_p(T)` for a
+   range of temperatures and a reference pressure :math:`p^\circ`;
+2. the molar enthalpy :math:`\hat{h}(T^\circ, p^\circ)` at :math:`p^\circ` and a reference
+   temperature :math:`T^\circ`;
+3. the absolute molar entropy :math:`\hat{s}(T^\circ, p^\circ)` at :math:`(T^\circ, p^\circ)`.
 
 See: :ref:`the Thermodynamic Models section <sec-thermo-models>`
```
bryanwweber, line 89: We probably don't need this link anymore since the very next section is the one being linked to 😄

pages/science/species-thermo.rst
```diff
@@ -80,25 +80,14 @@ expressions to compute the properties, they all require thermodynamic property
...
-1. the molar heat capacity at constant pressure :math:`\hat{c}^0_p(T)` for a
-   range of temperatures and a reference pressure :math:`P_0`;
-2. the molar enthalpy :math:`\hat{h}(T_0, P_0)` at :math:`P_0` and a reference
-   temperature :math:`T_0`;
-3. the absolute molar entropy :math:`\hat{s}(T_0, P_0)` at :math:`(T_0, P_0)`.
+1. the molar heat capacity at constant pressure :math:`\hat{c}^\circ_p(T)` for a
+   range of temperatures and a reference pressure :math:`p^\circ`;
+2. the molar enthalpy :math:`\hat{h}(T^\circ, p^\circ)` at :math:`p^\circ` and a reference
+   temperature :math:`T^\circ`;
+3. the absolute molar entropy :math:`\hat{s}(T^\circ, p^\circ)` at :math:`(T^\circ, p^\circ)`.
```
bryanwweber, line 87: Can you define what superscript-circle represents somewhere around here?

pages/science/species-thermo.rst
```diff
@@ -213,14 +202,14 @@ thermodynamic properties:
...
-   \hat{h}^0(T) = \hat{h}^0(T_0) + \hat{c}_p^0\cdot(T-T_0)
+   \hat{h}^\circ(T) = \hat{h}^\circ(T_0) + \hat{c}_p^\circ\cdot(T-T^\circ)
 
-   \hat{s}^0(T) = \hat{s}^0(T_0) + \hat{c}_p^0 \ln (T/T_0)
+   \hat{s}^\circ(T) = \hat{s}^\circ(T_0) + \hat{c}_p^\circ \ln (T/T^\circ)
 
-The parameterization uses four constants: :math:`T_0, \hat{c}_p^0(T_0),
-\hat{h}^0(T_0), \hat{s}^0(T)`. The default value of :math:`T_0` is 298.15 K; the
+The parameterization uses four constants: :math:`T^\circ, \hat{c}_p^\circ(T^\circ),
+\hat{h}^\circ(T^\circ), \hat{s}^\circ(T)`. The default value of :math:`T^\circ` is 298.15 K; the
```
bryanwweber, line 212:
```suggestion
\hat{h}^\circ(T^\circ), and \hat{s}^\circ(T)`. The default value of :math:`T^\circ` is 298.15 K; the
```

pages/science/species-thermo.rst
```diff
@@ -213,14 +202,14 @@ thermodynamic properties:
 
 .. math::
 
-   \hat{c}_p^0(T) = \hat{c}_p^0(T_0)
+   \hat{c}_p^\circ(T) = \hat{c}_p^\circ(T^\circ)
 
-   \hat{h}^0(T) = \hat{h}^0(T_0) + \hat{c}_p^0\cdot(T-T_0)
+   \hat{h}^\circ(T) = \hat{h}^\circ(T_0) + \hat{c}_p^\circ\cdot(T-T^\circ)
```
bryanwweber, line 207:
```suggestion
   \hat{h}^\circ(T) = \hat{h}^\circ\left(T_0\right) + \hat{c}_p^\circ \left(T-T^\circ\right)
```

pages/science/species-thermo.rst
```diff
@@ -213,14 +202,14 @@ thermodynamic properties:
...
 
-   \hat{s}^0(T) = \hat{s}^0(T_0) + \hat{c}_p^0 \ln (T/T_0)
+   \hat{s}^\circ(T) = \hat{s}^\circ(T_0) + \hat{c}_p^\circ \ln (T/T^\circ)
```
bryanwweber, line 209:
```suggestion
   \hat{s}^\circ(T) = \hat{s}^\circ(T_0) + \hat{c}_p^\circ \ln{\left(\frac{T}{T^\circ}\right)}
```

bryanwweber (approved):
Thanks for doing all this @decaluwe! The changes look great so far, just a few typo fixes and minor suggestions for clarity.

pages/science/species-thermo.rst
decaluwe: 😂

decaluwe:
Many thanks, @bryanwweber - just applied all the suggested changes, squashed the commits, and force-pushed.  Should be good to go!

bryanwweber (approved):

bryanwweber:
Wahoo in it goes!

decaluwe:
Awesome!  Now on to the next step... 😈

PR #183, opened 2022-03-27
Installation related updates for 2.6

speth:
Main updates:
- Add instructions for installing with pip
- Revise Windows instructions to account for removal of Python MSIs
- Update macOS Matlab install instructions
- Do CI builds for PRs that target the `testing` branch
- Fixes #193
- Clarify deprecation of CTI/XML
- Update recommended citation
- Fix some formatting issues with the guide for the C++ interface

PR #186, opened 2022-03-28
Also build when PR targets testing branch

ischoegl:
See subject - this allows checking of artifacts

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -7,9 +7,10 @@ on:
       - main
       - testing
   pull_request:
-    # Build when a pull request targets main
+    # Build when a pull request targets main or testing
     branches:
       - main
+      - testing
 
 env:
   NIKOLA_OUTPUT_DIR: output


bryanwweber (approved):

speth:
Doesn't this need to be on the testing branch? I had already included that change as part of #183.

bryanwweber:
@speth I think so? I think it possibly matters more what's in the file when the PR is created? I dunno.

ischoegl:
@speth … didn’t realize that you did the same on #183! From my perspective, I thought this needed to be on the `main` branch, unless CI is controlled by the targeted branch?

speth:
The CI looks at the config in the PR -- otherwise, you wouldn't be able to test changes to the CI configuration without pushing directly to `main` (or a different target branch).

ischoegl:
> The CI looks at the config in the PR

... not sure. Otherwise this PR would have triggered a build?

PR #184, opened 2022-03-27
Update instructions for Intel compilers and compilation on conda

ischoegl:
* [x] Update information on Intel compilers
* [x] Update information on compilation in `conda` environment

Closes #170, closes #181

PR #187, opened 2022-03-28
Fix indent in nested list

ischoegl:
Nested lists on reST require two character indents (not three) which created some mangled output in #184 ... added #186 so formatting can be checked also for PR's that target the `testing` branch.

### DIFF
--- a/pages/compiling/installation-reqs.rst
+++ b/pages/compiling/installation-reqs.rst
@@ -53,6 +53,11 @@ General Notes
 * You will need to install compilers for your system by following the instructions in the sections
   below to install the compiler for your operating system.
 
+* By default, Cantera is installed into the active conda environment, where the
+  layout of the directory structure corresponds to the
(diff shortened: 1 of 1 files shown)


pages/compiling/installation-reqs.rst
```diff
@@ -154,18 +159,10 @@ Conda Requirements
...
-  .. code:: bash
-
-     scons build
-     scons test
-
-* To install Cantera, use the command
-
-  .. code:: bash
-
-     scons install
```
bryanwweber, line 168: Any reason to remove this?

ischoegl: Yes. If you look at the actual page, the next step is to download the source :wink: ... https://testing.cantera.org/compiling/installation-reqs.html#conda-requirements

bryanwweber (approved):

PR #188, opened 2022-03-28
Add information on testing to README.md

### LINKED ISSUE #88: How should we handle updates to website content for the development version?

ischoegl:
Add information (and link) for [testing.cantera.org](https://testing.cantera.org/) to README

Addresses #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,10 @@
 This repository contains the source code for the [cantera.org](https://cantera.org) website. The
 site is built using the [Nikola](https://getnikola.com) static site generator.
 
+While the state of the website corresponds to the `main` branch, upcoming changes can
+be previewed on [testing.cantera.org](https://testing.cantera.org), which is based on
+the `testing` branch of the repository.
+
 ## To build the website
 
 1. Create a virtual environment for Nikola (i.e., using conda or virtualenv)


bryanwweber:
I don't particularly want to make this public, there's no reason for the general public to go there and I don't want Google to get confused.

ischoegl:
Np. Nevertheless, I hope to resolve #88 before long (after 2.6) as updates to the website target an active branch while referring to the stable version ... ideally, we have a stable official website and a repository that collects updates for the next release.

bryanwweber:
Thanks @ischoegl! Agreed, I hope we can resolve #88 as well. I don't think the URL should be `testing.cantera.org` though, and we'll have to get creative with HTML headers to make sure Google doesn't index the wrong data.

Issue #175, opened 2022-02-14
Reformatting Reactors and Reactor Networks science page for clarity


chinahg:
Current Reactors and Reactor Networks science page is very long and does not provide clear information on which governing equations correspond to each reactor type. 

Proposing formatting changes to simplify user interface. Possibly separate pages for each reactor type including delegated reactors (using reactor science page as a home base for hyperlinks). Reactor networks will also have a separate, linked, page that includes integration documentation. Walls, surface reactions, and flow devices will remain on initial Reactor science page. Ideally also include diagrams to clarify what each variable in the governing equations are.

speth:
I think this is done now that #189 has been merged.

PR #189, opened 2022-03-31
Reformatted Reactors homepage, in progress

chinahg:
Addressing open issue #175


speth:
@chinahg - Rather than merging the `main` branch into your working branch, can you revert that and rebase onto `main` instead?

chinahg:
> @chinahg - Rather than merging the `main` branch into your working branch, can you revert that and rebase onto `main` instead?

Done, let me know if there is anything that is out of place.

pages/science/reactors/customreactor.rst
```diff
@@ -0,0 +1,45 @@
...
+for solving problems where Cantera's reactor network model cannot 
+be configured to describe the system in question, as in a `Delegated
+Reactor </science/reactors/delegatedreactor.html>`__.
+
+The process to bypass Cantera's existing governing equations can be
+described in 7 steps:
+
+#. Define your new class
+   
+   a. Create _init_ function
```
speth, line 32: Watch the formatting here.

pages/science/reactors/customreactor.rst
```diff
@@ -0,0 +1,45 @@
...
+   b. Specify the new custom ODEs
+
+#. Create a new gas object and set initial conditions
+#. Integrate custom equations over a single timestep
+#. Set new gas state to the values obtained by the supplied ODE solver
+#. Save the gas state
+#. Repeat integration over all timesteps (steps 3-5)
+#. The final state vector contains all gas properties obtained from Cantera for each timestep
```
speth, line 40: This doesn't really make sense as a description of how the ODEs are solved. I think we need to iterate a bit on this to make it more useful and technically correct.

pages/science/reactors/reactors.rst
```diff
@@ -0,0 +1,482 @@
...
+
+Time Integration for Reactor Networks
+-------------------------------------
+
+Cantera uses the CVODES solver from the `SUNDIALS 
+<https://computing.llnl.gov/projects/sundials>`__ 
+package to integrate the stiff ODEs of reacting systems. These stiff ODEs are referring to
+the governing equations defining the reactors above.
+More in-depth information on the CVODES solver can be found `here </science/reactors
+/cvodes.html>`__.
```
speth, line 237: I think we discussed this, but looking at the page it's pretty clear that this doesn't call enough attention to the fact that there's a whole additional page of info here -- I think adding a card for CVODES would make a lot of sense.

speth (changes requested):
Thanks for starting this, @chinahg. I think the structural changes here make a lot of sense.

I think what you've done with pulling some of the documentation from @paulblum's blog post on how CVODES and ReactorNet work makes a lot of sense, but this section could use some more work. In part, some of the formatting and linking of content needs to be updated since those posts were in Markdown while this new file is in RST -- compare the HTML output for each and you'll see the differences. I also think that this needs to be cleaned up a bit if it's going to be part of the permanent documentation -- for example, I don't think links to specific line numbers or source files in the code make much sense, as these will change over time. Linking to specific classes and methods in the Doxygen / Sphinx docs does make sense, though.

pages/science/reactors/reactors.rst
```diff
@@ -0,0 +1,482 @@
...
+      .. container:: col-12
+
+         .. container:: card-deck
+
+            .. container:: card
+
+               .. container::
+                  :tagname: a
+                  :attributes: href="controlreactor.html"
+                              title="Control Volume Reactor"
```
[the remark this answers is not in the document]
chinahg: yes, I forgot to track the new file. Added.

speth:
@chinahg, can you rebase this on the current `main` branch and resolve the conflicts?

bryanwweber:
Yes, it moved to a submodule. You can run ` git submodule update --init --recursive` to pull them in.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+      This guide explains ways Cantera can solve the governing equations of 
+      a transient Reactor Network problem. Additional insights 
+      into Cantera's integrator (CVODES) are also provided.
+
+Using Cantera to Advance a Reactor Network in Time
+**************************************************
+
+Starting off the current state of the system, it can be advanced in time by 
+one of the following methods:
```
bryanwweber, line 20:
```suggestion
A ``ReactorNetwork`` can be advanced in time by one of the following three methods:
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+- ``step()``: The step method computes the state of the system at the a priori
+  unspecified time :math:`t_{\mathrm{new}}`. The time :math:`t_{\mathrm{new}}`
+  is internally computed so that all states of the system only change within a
+  (specifiable) band of absolute and relative tolerances. Additionally, the time
+  step must not be larger than a predefined maximum time step
+  :math:`\Delta t_{\mathrm{max}}`. The new time :math:`t_{\mathrm{new}}` is
+  returned by this function.
```
bryanwweber, line 28:
```suggestion
- ``step()``: The ``step()`` method computes the state of the system after one time step. The size of the step is determined by CVODES when the method is called. CVODES determines the step size by <fill in half a sentence of detail>. The time step must not be larger than a predefined maximum time step :math:`\Delta t_{\mathrm{max}}`. The new time :math:`t_{\mathrm{new}}` at the end of the single step is returned by this function.This method produces the highest time resolution in the output data of the methods implemented in Cantera.
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+- ``advance(``\ :math:`t_{\mathrm{new}}`\ ``)``: This method computes the state of the
+  system at time :math:`t_{\mathrm{new}}`. :math:`t_{\mathrm{new}}` describes
+  the absolute time from the initial time of the system. By calling this method
+  in a for loop for pre-defined times, the state of the system is obtained for
+  exactly the times specified. Internally, several ``step()`` calls are
+  typically performed to reach the accurate state at time
+  :math:`t_{\mathrm{new}}`.
```
bryanwweber, line 36:
```suggestion
- ``advance(t_new)``: This method computes the state of the system at the user-provided time :math:`t_{\mathrm{new}}`. :math:`t_{\mathrm{new}}` is the absolute time from the initial time of the system. Although the user specifies the time when integration should stop, CVODES chooses the time step size as the network is integrated. Many of these internal CVODES time steps are usually required to reach :math:`t_{\mathrm{new}}`. As such, ``advance(t_new)`` preserves the accuracy of using ``step()`` but allows consistent time step spacing in the output data.
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+  reactor network is of interest, this method can be used. Internally, the
+  steady state is approached by time stepping. The network is considered to be
+  at steady state if the feature-scaled residual of the state vector is below a
+  given threshold value (which by default is 10 times the time step ``rtol``).
+
+The use of the ``advance`` method in a loop has the advantage that it produces
+results corresponding to a predefined time series. These are associated with a
+predefined memory consumption and well comparable between simulation runs with
+different parameters. However, some detail (for example, a fast ignition process)
+might not be resolved in the output data due to the typically large time steps.
```
bryanwweber, line 49:
```suggestion
The ``advance(t_new)`` is typically used when consistent, regular, time steps are required in the output data. This usually simplifies comparisons among many simulation results at a single time point. However, some detail, for example, a fast ignition process, might not be resolved in the output data if the user-provided time step is not small enough.

```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+To avoid losing this detail, the
+`Reactor::setAdvanceLimit <{{% ct_docs doxygen/html/dc/d5e/classCantera_1_1Reactor.html#a9b630edc7d836e901886d7fd81134d9e %}}>`__
+method (C++) or the :py:func:`Reactor.set_advance_limit` method (Python) can be
+used to set the maximum amount that a specified solution component can change
+between output times. For an example of this feature's use, see the example
+`reactor1.py </examples/python/reactors/reactor1.py.html>`__.
```
bryanwweber, line 55:
```suggestion
`reactor1.py </examples/python/reactors/reactor1.py.html>`__. However, as a tradeoff, the time step sizes in the output data are no longer guaranteed to be uniform.
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+The ``step`` method results in many more data points because of the small
+timesteps needed. Additionally, the absolute time has to be kept track of
+manually.
```
bryanwweber, line 59:
```suggestion
The ``step()`` method results in many more data points because of the small
timesteps needed. Additionally, the absolute time has to be kept track of
manually.
```

I'm not sure what the second sentence means. The absolute time is available from `ReactorNet.time`, and you have to keep track of the time when you're using `advance(t_new)` as well... I suppose you could have a list of times to fill in...

```python
times = np.arange(1, 10)
for t in times:
    netw.advance(t)
```
but you could just as easily compute the new time by `t + dt` and have to add it to a list anyways. So I'd suggest deleting this paragraph, I don't think it adds much considering the change I suggested above for the `step()` description.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+Even though Cantera comes pre-defined with typical parameters for tolerances
+and the maximum internal time step, the solution sometimes diverges. To solve
```
bryanwweber, line 62: What does `diverges` mean in this context?

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+this problem, three parameters can be tuned: The absolute time stepping
+tolerances, the relative time stepping tolerances, and the maximum time step. A
+reduction of the latter value is particularly useful when dealing with abrupt
+changes in the boundary conditions (for example, opening/closing valves; see
+also the `IC engine example </examples/python/reactors/ic_engine.py.html>`__).
```
bryanwweber, line 67: Not only is it particularly useful, it's practically required in this case. But on the other side, it's also pretty much the only case when you want to set this parameter. Otherwise you're introducing unnecessary stiffness into the solution by forcing more time steps that aren't needed.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+
+How Does Cantera's Reactor Network Time Integration Feature Actually Work?
+==========================================================================
+
+There's a great description of the science behind Cantera's reactor network 
+simulation capabilities available on the Cantera website, 
+`here <https://cantera.org/science/reactors/reactors.html>`__. This section will go into more 
+developer-oriented detail about how the last step, `ReactorNet`'s 
+`time integration methods <https://cantera.org/science/reactors/reactors.html#time-
+integration-for-reactor-networks>`__, actually work. A `ReactorNet` object doesn't
```
bryanwweber, line 77:
```suggestion
integration-for-reactor-networks>`__, actually work. A ``ReactorNet`` object doesn't 
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+How Does Cantera's Reactor Network Time Integration Feature Actually Work?
+==========================================================================
+
+There's a great description of the science behind Cantera's reactor network 
+simulation capabilities available on the Cantera website, 
+`here <https://cantera.org/science/reactors/reactors.html>`__. This section will go into more 
+developer-oriented detail about how the last step, `ReactorNet`'s 
+`time integration methods <https://cantera.org/science/reactors/reactors.html#time-
+integration-for-reactor-networks>`__, actually work. A `ReactorNet` object doesn't 
+perform time integration on its own. It generates a system of ODE's based on the
```
bryanwweber, line 78:
```suggestion
perform time integration on its own. It generates a system of ODEs based on the 
```

What does it mean to "generate" a system of ODEs in this context? Cantera never actually writes down any of the equations from the `reactors.html` page... It's probably more accurate to say something more like, Cantera generates a Jacobian array for the set of Reactors installed in the ReactorNet. This Jacobian, along with functions to evaluate the residual of each ODE representing the system, is passed to an Integrator. You'll need to tweak the phrasing there probably, but I think that's somewhat better 😄

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,242 @@
...
+from `FuncEval` will implement the evaluation of the provided ODE system.
+
+An ODE right-hand-side evaluator is always needed in the ODE solution process (it's the only way to describe the system!), and for that reason a `FuncEval` object is a required parameter 
+when initializing any type of `Integrator`.
+
+Let's take a look at how `ReactorNet` implements this `FuncEval` object. `ReactorNet` actually points to itself when 
+defining a `FuncEval` type, meaning it defines *itself* as a `FuncEval` derivative.
+
+Then, `ReactorNet` initializes the `Integrator`, using a reference to itself (as a `FuncEval`) from the 
+`this <https://www.geeksforgeeks.org/this-pointer-in-c/>`__ pointer.
```
bryanwweber, line 195: Can we find a better reference than `geeksforgeeks`? I find their site nearly unusable with all the ads, popups to subscribe, etc.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,160 @@
...
+in describing a certain configuration. In this situation, Cantera 
+Reactors' governing equations can be modified to the user's specific 
+needs. An Extensible Reactor allows modifications of a Reactor class' 
+`governing equations </science/reactors/reactors.html>`__. If the governing equations 
+must be entirely replaced or an integrator besides CVODES is desired, a 
+`Custom Reactor </science/reactors/customreactor.html>`__
+may be a more appropriate Reactor class to use.
+
+The state variables available for modification 
+depend on the type of Reactor base class chosen. For example, choosing
```
bryanwweber, line 27: In the last paragraph, you said the governing equations are modified. Here, you're saying the state variables can be  modified. These are separate ideas, so we should clarify.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,160 @@
...
+needs. An Extensible Reactor allows modifications of a Reactor class' 
+`governing equations </science/reactors/reactors.html>`__. If the governing equations 
+must be entirely replaced or an integrator besides CVODES is desired, a 
+`Custom Reactor </science/reactors/customreactor.html>`__
+may be a more appropriate Reactor class to use.
+
+The state variables available for modification 
+depend on the type of Reactor base class chosen. For example, choosing 
+an `Ideal Gas Constant Pressure Reactor </science/reactors/
+idealgasconstpresreactor.html>`__ allows
```
bryanwweber, line 29: I'd suggest using a `:doc:` or `:ref:` style reference here, to refer to the document or section you want to link to, instead of the absolute URL.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,160 @@
...
+   cp_lump = 1.0 #[J/kgK]
+
+   #initialize time at zero
+   time = 0 #[s]
+   n_steps = 300
+
+2. Define a new custom Reactor class. Here we named it "DummyReactor" and 
+chose the Ideal Gas Constant Pressure Reactor as the base class to inheret
+governing equations from. An Extensible Reactor is also known as a Delegated 
+Reactor.
```
bryanwweber, line 122: This last bit about extensible vs. delegated needs to be highlighted in a more prominent way, IMO. I missed it on first reading this page.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,160 @@
...
+
+   class DummyReactor(ct.DelegatedIdealGasConstPressureReactor):
+
+      #modify energy equation to include solid mass in reactor
+
+      def after_eval(self, t, LHS, RHS):
+
+         self.m_mass = mass_gas
+
+         LHS[1] = mass_rock*cp_rock+self.m_mass*self.thermo.cp_mass
```
bryanwweber, line 136:
```suggestion
         LHS[1] = mass_rock * cp_rock + self.m_mass * self.thermo.cp_mass
```
`mass_rock`, `cp_rock`, `Q`, and `mass_gas` are all undefined, you'll get a `NameError` if you actually try to run this code.

bryanwweber (dismissed):
Hi @chinahg! I'm about half-way through a review of these changes, thanks for taking on this big project! I wanted to send you the comments I have so far, because some of them might need a bit of discussion that we can get started now 😄 This looks really great overall and I mostly have nitpick comments about phrasing and consistency

pages/science/reactors/extensiblereactor.rst
chinahg: I have those variables defined previously in the example, here's the full script the documentatin is based on that I was trying to document.
`    def test_RHS_LHS(self):
        #create gas object
        gas = ct.Solution('h2o2.yaml')
        gas.TPX = 500, ct.one_atm, 'H2:2,O2:1,N2:4'
        gas_initial_enthalpy = gas.enthalpy_mass

        #define properties of gas and solid
        mass_gas = 20 #[kg]
        Q = 100 #[J/s]
        mass_lump = 10 #[kg]
        cp_lump = 1.0 #[J/kgK]

        #initialize time at zero
        time = 0 #[s]
        n_steps = 300

        #define a class representing reactor with a solid mass and
        #gas inside of it
        class DummyReactor(ct.DelegatedIdealGasConstPressureReactor):
            #modify energy equation to include solid mass in reactor
            def after_eval(self,t,LHS,RHS):
                self.m_mass = mass_gas
                LHS[1] = mass_lump*cp_lump+self.m_mass*self.thermo.cp_mass
                RHS[1] = Q

        r1 = DummyReactor(gas)
        r1_net = ct.ReactorNet([r1])

        for n in range(n_steps):
            time += 4.e-4
            r1_net.advance(time)`

bryanwweber: Apologies if I missed it, but I don't see those variables anywhere else on the page? I'm also not a big fan of splitting a script over multiple code blocks... It's harder to follow as a reader, especially if you're only skimming, and I can't easily copy-paste the parts of the example that might be relevant for my problem, since I have to replace these variables or remove them or rename them

chinahg: Ah I see, there was a change in naming that didn't get updated. `mass_lump` and `lump_cp` are `rock_mass` and `rock_cp` defined on lines 112 and 113. `Q` and `mass_gas` are defined on line 110 and 111. I agree about splitting code, maybe I'll just include the steps as comments in a single code block instead.

pages/science/reactors/extensiblereactor.rst
chinahg: I'm not sure how to use `:ref:` to reference across `.rst` files... I poked around and the examples/documentation was for referencing headers within the file.

bryanwweber: No problem, we can fix it later.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+
+   #create gas object
+   gas = ct.Solution('h2o2.yaml')
+   gas.TPX = 500, ct.one_atm, 'H2:2,O2:1,N2:4'
+
+   #define properties of gas and solid
+   mass_gas = 20 #[kg]
+   Q = 100 #[J/s]
+   mass_rock = 10 #[kg]
+   cp_rock = 1.0 #[J/kgK]
```
speth, line 121: This is kind of pedantic, but it's worth it for code that we supply as examples. PEP8 suggests:
- Each line of a block comment starts with a # and a single space (unless it is indented text inside the comment).
- Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.
So this should be:
```suggestion
   # define properties of gas and solid
   mass_gas = 20  # [kg]
   Q = 100  # [J/s]
   mass_rock = 10  # [kg]
   cp_rock = 1.0  # [J/kgK]
```

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+
+   #initialize time at zero
+   time = 0 #[s]
+   n_steps = 300
+
+  #2 Define a new custom Reactor class. Here we named it "DummyReactor" and
```
speth, line 127:
```suggestion
  #2 Define a new custom Reactor class. Here we named it "RockReactor" and 
```

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+   #initialize time at zero
+   time = 0 #[s]
+   n_steps = 300
+
+  #2 Define a new custom Reactor class. Here we named it "DummyReactor" and 
+   #chose the Ideal Gas Constant Pressure Reactor as the base class to inheret
+   #governing equations from. 
+
+   #define a class representing reactor with a solid mass and gas inside of it
+   class RockReactor(ct.DelegatedIdealGasConstPressureReactor):
```
speth, line 132:
```suggestion
   class RockReactor(ct.ExtensibleIdealGasConstPressureReactor):
```

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+
+      #modify energy equation to include solid mass in reactor
+      
+      #after the initial solution for time t is computed ask Cantera to solve the modified 
+      #equation. The index 1 refers to modification of governing equation 2 in the reactor
+      #documentation (recall that indexing begins at 0).
+
+      def after_eval(self, t, LHS, RHS):
+      #although the time variable t is not used directly in the method definition it is a 
+      #required argument for the internal solver.
```
speth, line 142: The comment here should be indented like the body of the function, not flush with the `def`.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+The final state vector for your reactor network contains the final gas 
+properties obtained from Cantera using the modified equation(s).
+
+Details on functions in addition to ``eval()`` 
+that are able to be modified with **before_**, **after_**, or 
+**replace_** can be found `here 
+<https://cantera.org/documentation/dev/sphinx/html/cython/zerodim.html?highlight=
+extensible#extensiblereactor>`__.
+
+An Extensible Reactor is also known as a Delegated Reactor.
```
speth, line 169: I don't think the "Delegated Reactor" term is used anywhere in the final implementation -- the only part that still bears this name is the `Delegator` class which is used to provide the `ExtensibleReactor` classes.

pages/science/reactors/extensiblereactor.rst
```diff
@@ -0,0 +1,181 @@
...
+
+
+
+
+
+More in-depth documentation on the different ways to modify equations using
+an Extensible Reactor can be found `here <https://www.cantera.org/documentation
+/dev/doxygen/html/de/d7e/classCantera_1_1ReactorDelegator.html>`__ and `here 
+<https://cantera.org/documentation/dev/sphinx/html/cython/zerodim.html?highlight=
+extensible#extensiblereactor>`__.
```
speth, line 181: Since 2.6.0 is out, it shouldn't be necessary to link to the dev docs, and you can link to the version-independent stable docs instead.

pages/science/reactors/reactors.rst
```diff
@@ -0,0 +1,520 @@
...
+difference. Interactions of the reactor with the environment are defined on one or more *walls*,
+*inlets*, and *outlets*.
+
+In addition to single reactors, Cantera is also able to interconnect reactors into a Reactor
+Network. Each reactor in a network may be connected so that the contents of one reactor flow into
+another. Reactors may also be in contact with one another or the environment via walls that conduct
+heat or move to do work.
+
+Reactor Types and Governing Equations
+=======================================
```
speth, line 53: The underline should be the same length as the heading.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Using Cantera to Advance a Reactor Network in Time
+**************************************************
+
+A ``ReactorNetwork`` can be advanced in time by one of the following three 
+methods:
+
+- ``step()``: The ``step()`` method computes the state of the system after one 
+  time step. The size of the step is determined by CVODES when the method is called. 
+  CVODES determines the step size by <fill in half a sentence of detail>. The time
```
speth, line 24: This still needs some work...

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+
+A ``ReactorNetwork`` can be advanced in time by one of the following three 
+methods:
+
+- ``step()``: The ``step()`` method computes the state of the system after one 
+  time step. The size of the step is determined by CVODES when the method is called. 
+  CVODES determines the step size by <fill in half a sentence of detail>. The time 
+  step must not be larger than a predefined maximum time step 
+  :math:`\Delta t_{\mathrm{max}}`. The new time :math:`t_{\mathrm{new}}` at the end 
+  of the single step is returned by this function.This method produces the highest time
```
speth, line 27:
```suggestion
  of the single step is returned by this function. This method produces the highest time 
```

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+when dealing with abrupt changes in the boundary conditions (for example, 
+opening/closing valves; see also the `IC engine example </examples/python/reactors
+/ic_engine.py.html>`__).
+
+How Does Cantera's Reactor Network Time Integration Feature Actually Work?
+==========================================================================
+
+There's a great description of the science behind Cantera's reactor network 
+simulation capabilities available on the Cantera website, 
+`here <https://cantera.org/science/reactors/reactors.html>`__. This section will go into more
```
speth, line 74: I'm not sure we need to pat ourselves on the back about how *great* the other page is...

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+First, let's take a look at a basic example to see how we might utilize Cantera's time integration 
+functionality. We'll simulate an isolated reactor in Python that is homogeneously filled by a gas 
+mixture (the gas state used in this example is arbitrary, but interesting because it's 
+explosive). Then we'll advance the simulation in time to an (arbitrary) absolute time of 
+1 second, noting the changes in the state of the gas.
+
+.. code-block::
+
+    >>> import cantera as ct                           #import Cantera's Python module
+    >>> gas = ct.Solution("gri30.yaml")                #create a default GRI-Mech 3.0 gas mixture
```
speth, line 98: I'd ditch the `>>>` prefixes, since this isn't intermingling inputs and outputs, and make sure that this gets formatted nicely with the correct syntax highlighting in the HTML output.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+pointer to an ``Integrator`` instance of type ``itype``. The ``newIntegrator()`` instance will automatically 
+have an ``itype`` of ``CVODES``, which is installed with Cantera. The ``newIntegrator()`` will be stored as 
+variable ``m_integ``.
+
+Step 4: Communicate with CVODES using a wrapper function
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Because ``CVODES`` is written in C, the ``CVodesIntegrator`` C++ wrapper is used to access the solver.
+The ``CVodesIntegrator`` class is a C++ wrapper class for ``CVODES``. (`Documentation 
+<https://cantera.org/documentation/docs-2.4/doxygen/html/d9/d6b/classCantera_1_1CVodesIntegrator.html>`__)
```
speth, line 140: Let's definitely not link to the Cantera 2.4 documentation -- please use the version-independent way of doing this.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+``FuncEval`` an abstract class:
+
+To evaluate the reactor governing equations the following parameters must be known:
+
+@param[in] t time.
+@param[out] LHS pointer to start of vector of left-hand side 
+coefficients for governing equations, length m_nv, default values 1
+@param[out] RHS pointer to start of vector of right-hand side 
+coefficients for governing equations, length m_nv, default values 0
+
```
speth, line 211: This needs some formatting work -- it currently just all gets written out as if it was a paragraph.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+To evaluate the reactor governing equations the following parameters must be known:
+
+@param[in] t time.
+@param[out] LHS pointer to start of vector of left-hand side 
+coefficients for governing equations, length m_nv, default values 1
+@param[out] RHS pointer to start of vector of right-hand side 
+coefficients for governing equations, length m_nv, default values 0
+    
+.. code-block::
+    virtual void eval(double t, double* LHS, double* RHS);
```
speth, line 213: a `code-block` needs a blank line before the start of the code -- currently this is generating an angry red error message in the HTML output.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+state varaible to the solution vector (``y``).
+
+Step 7: ``eval()`` is called to solve provided ODEs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Along with the rest of ``FuncEval``'s virtual functions, an appropriate override is provided for ``FuncEval::eval()`` in 
+``ReactorNet``
+
+.. code-block::
+
```
speth, line 228: You should tell this that it's C++ code so the following code will be formatted nicely.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+    void ReactorNet::eval(doublereal t, doublereal* y, doublereal* ydot, doublereal* p)
+    {
+        m_time = t; // This will be replaced at the end of the timestep
+        updateState(y);
+        for (size_t n = 0; n < m_reactors.size(); n++) {
+            m_reactors[n]->evalEqs(t, y + m_start[n], ydot + m_start[n], p);
+        }
+        checkFinite("ydot", ydot, m_nv);
```
speth, line 236: Well, the `evalEqs` method no longer exists, so this is a bit out of date...

pages/science/reactors/customreactor.rst
```diff
@@ -0,0 +1,49 @@
...
+
+      This page defines and links to an example of a Custom Reactor.
+
+Custom Reactors
+***************
+
+If the governing 
+equations must be entirely replaced or an integrator besides CVODES is 
+desired, a `Custom Reactor </science/reactors/customreactor.html>`__
+may be the most appropriate Reactor class to use. In this situation,
```
speth, line 20: To help distinguish between this approach and the `ExtensibleReactor` option, maybe we should just refer to this as "using an external integrator"? At this point, I would hope that the `ExtensibleReactor` approach covers most use cases and that people shouldn't need to fall back to this option very often.

pages/science/reactors/cvodes.rst
```diff
@@ -0,0 +1,244 @@
...
+    >>> gas.TPX = 1000.0, ct.one_atm, "H2:2,O2:1,N2:4" #set gas to an interesting state
+    >>> reac = ct.IdealGasReactor(gas)                 #create a reactor containing the gas
+    >>> sim = ct.ReactorNet([reac])                    #add the reactor to a new ReactorNet simulator
+    >>> gas()                #view the initial state of the mixture (state summary will be printed to console)
+    >>> sim.advance(1)       #advance the simulation to the specified absolute time, t = 1 sec
+    >>> gas()                #view the updated state of the mixture, reflecting properties at t = 1 sec
+
+For a more advanced example that adds inlets and outlets to the reactor, see Cantera's combustor example 
+(`Python <https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/examples/reactors/combustor.py>`__ 
+| `C++ <https://github.com/Cantera/cantera/blob/main/samples/cxx/combustor/combustor.cpp>`__). Additional examples
```
speth, line 108: Better to link to these directly on the Cantera website as `/examples/python/reactors/combustor.py.html` and `/examples/cxx/combustor.html`.

speth (changes requested):
Thanks for these updates, @chinahg. I think the biggest thing to do here is to go through here and make sure that all of the links being referenced are up to date and written in a way that won't require us to keep updating them. There is perhaps some room for additional wordsmithing, but I think we can do that in additional PRs.

bryanwweber:
@chinahg I agree with @speth let's push this across the finish line!

speth (approved):

bryanwweber:
Merged, congrats @chinahg!

PR #192, opened 2022-04-04
Update C++ tutorial

### LINKED ISSUE #191: Update very simple C++ program
The [very simple C++ program](https://cantera.org/tutorials/cxx-guide/simple-example.html) should be updated to use `newSolution`.

ischoegl:
- [x] Replace `newPhase` by `newSolution` in C++ examples
- [x] Add information on environment variables to C++ compilation tutorial

Closes #191

### DIFF
--- a/pages/tutorials/cxx-guide/compiling.rst
+++ b/pages/tutorials/cxx-guide/compiling.rst
@@ -18,6 +18,17 @@ specifying the appropriate header and library paths, and specifying the required
 libraries when linking. It is also necessary to specify the paths for libraries
 used by Cantera, such as Sundials, BLAS, and LAPACK.
 
+Instructions below assume a Linux operating system where Cantera's libraries are
+installed in a standard location such as ``/usr/lib`` or ``/usr/local/lib``. If Cantera
+is installed to a custom location, environment variables ``LD_LIBRARY_PATH`` and
+``PKG_CONFIG_PATH`` need to be specified. If they are not already set elsewhere, they
+can be set from the command line as:
+
+.. code:: bash
+
+   export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/lib
+   export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/path/to/lib/pkgconfig
+
 pkg-config
 ==========
 
@@ -67,7 +78,7 @@ look like this:
               CPPPATH=['/usr/local/cantera/include',
                        '/usr/local/sundials/include'],
               LIBS=['cantera', 'sundials_cvodes', 'sundials_ida',
-                    'sundials_nvecserial', 'lapack', 'blas'],
+                    'sundials_nvecserial', 'lapack', 'blas', 'fmt', 'yaml-cpp'],
               LIBPATH=['/usr/local/cantera/lib',
                        '/usr/local/sundials/lib'],
               LINKFLAGS=['-g', '-pthread'])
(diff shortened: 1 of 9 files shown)


pages/tutorials/cxx-guide/compiling.rst
```diff
@@ -18,6 +18,19 @@ specifying the appropriate header and library paths, and specifying the required
...
+from the command line as:
+
+.. code:: bash
+
+   LD_LIBRARY_PATH=/path/to/lib
+   PKG_CONFIG_PATH=/path/to/lib/pkgconfig
+
+Note that when using *conda*, environment variables need to be specified as
+``conda env config vars set name=value`` (see `conda user guide
+<https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables>`__.
```
speth, line 32: I think there are couple of things getting mixed up here. To set _environment_ variables that affect subsequent commands, you need to run `export PKG_CONFIG_PATH=/path/to/lib/pkgconfig` for example. What's shown above is just defining a local variable.

The suggestion for conda here is something else -- that's a way of specifying environment variables that will then be set whenever the environment is activated. This might make an interesting addition to our conda package, if that's something that can be incorporated there, but I'm not sure how relevant it is here.

ischoegl: :joy: ... yup, this explains a couple of things! This just shows how much I relied on `setup_cantera` in the past as I was pretty much able to never define any variables on the CLI. Had thought that `export` was only necessary from within a file. Oh well.

Fair point about the *conda* pathway. I can't really incorporate this here as this PR targets `main` and some of the new conda changes point to `testing`.

ischoegl:
@speth ... ended up replacing `newPhase` in the remaining tutorials also, and fixed a couple of other things that were not up-to-date in the documentation. I tested compilation of all examples, and things work as expected.

ischoegl:
Fwiw, some of the includes can be simplified if Cantera/cantera#1238 were to be adopted.

bryanwweber:
As @speth mentioned elsewhere, I don't think now is a good time to get into a conversation about the content of the include files. I'd prefer not dropping the lowercase include files from the examples.

I think we should limit this PR to the changes in the `export` and library configuration, and defer changing the examples until later. Since the website can be updated independently of a release, we can change these to `newSolution` later.

ischoegl:
@bryanwweber ... thank you for the comments.

> As @speth mentioned elsewhere, I don't think now is a good time to get into a conversation about the content of the include files. I'd prefer not dropping the lowercase include files from the examples.

That's totally fine. At the same time, not having `Solution.h` covered by the convenience headers is not great.

> I think we should limit this PR to the changes in the export and library configuration, and defer changing the examples until later. Since the website can be updated independently of a release, we can change these to `newSolution` later.

I'm not sure I agree here: `newSolution` pretty much replaced most of the instances in the C++ test suite already, and is overall much more convenient for the users to use (see `factory_demo.cpp`!). 

My preferred solution would be to switch the imports to
```C++
#include "cantera/core.h"
```
as proposed in PR Cantera/cantera#1238

ischoegl:
@bryanwweber and @speth ... now that `cantera/core.h` is deferred, I updated the examples to be up to date for the 2.6 release. This is ready for another round of reviews.

speth (approved):
Thanks, @ischoegl. This looks good to me.

PR #194, opened 2022-04-05
Fix inline literals in HTML title of CXX guide

### LINKED ISSUE #193: Inline literals not rendered properly in titles
When used in titles, some ``` ``inline literals`` ``` are not rendered properly. For an example, see [here](https://cantera.org/tutorials/cxx-guide/factories.html)

bryanwweber:
Closes #193

### DIFF
--- a/pages/tutorials/cxx-guide/factories.rst
+++ b/pages/tutorials/cxx-guide/factories.rst
@@ -4,7 +4,7 @@
 
    .. raw:: html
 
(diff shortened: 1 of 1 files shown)


speth:
I had noticed this and made a different fix in c7e1ea22 (#183) where I just removed the formatting for these class names. I noticed that the change here leads to these names being printed in black text, which is sort of jarring in this context where you end up with every other word being blue or black. Would it be better to just format these with `<tt>`?

bryanwweber:
@speth I think just removing the formatting altogether is a fine fix. I'll close this.

PR #219, opened 2022-08-07
Update Python example location

ischoegl:
As Python examples are no longer packaged with the Python module after Cantera/cantera#1352, the website needs to be updated accordingly.


bryanwweber (approved):
Thanks for updating this!

PR #237, opened 2023-03-17
Add Python RGFROSH to affiliated

corykinney:
Hey! I was hoping to add the package [VasuLab/RGFROSH](https://github.com/VasuLab/RGFROSH) to the affiliated packages page of the Cantera website. 

It is a solver for shock conditions in a shock tube for arbitrary equations of state - with a focus on real gases. The package itself can be used with any backend, but its interface was specifically designed for Cantera so that `ct.ThermoPhase` objects can be passed directly to the solver.


bryanwweber (approved):
LGTM!

PR #260, opened 2023-11-12
Fix density formula and momentum equation

imitrichev:
Fix density formula rho=(PM)/(RT)

and momentum equation (A/V is already taken into account in \dot m_s

Also, please provide a description for \dot w_k, I don't understand what is it and if it is really needed in conservatin eq. for fluid species k.

### DIFF
--- a/pages/science/reactors/pfr.rst
+++ b/pages/science/reactors/pfr.rst
@@ -83,7 +83,7 @@ overall mass conservation equation for the reactor can then be written as:
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


speth (approved):
Thanks, these changes make sense to me.

$\dot \omega_k$ is the molar rate of production of species $k$ from gas phase reactions, which definitely does need to appear in the species and energy equations. $\dot s_k$ in contrast is the molar rate of production from surface reactions. These are the same definitions used in the other reactor formulations (though the definitions should probably be given again here).

imitrichev:
Dear Ray,
thank you for approving this.

Also, thank for explaining about \dot w - it is about gas-phase reactions. Of course, if PFR can account for gas-phase chemistry, these terms should be present in the equations.

PR #263, opened 2024-03-07
Convert website to Sphinx

### LINKED ISSUE #209: Change HTML generator to Sphinx
Since this website was redesigned in 2018, the tradeoffs that led us to choose Nikola at that time have shifted in favor of using Sphinx. This issue will track the progress of moving the HTML generation of the `cantera-website` repository to use Sphinx instead of Nikola.

## Action Items

- [x] Use Sphinx `conf.py` format
- [ ] Ensure examples can be generated
- [ ] Move content files as needed to support Sphinx's directory layout

## Non-Goals

- Modify or add any content. It may be necessary to move content files around to support Sphinx, but we do not intend this to be significant modifications to the content, or structure of the content, of the site.
- Add plugins or parsers to support content that is not already on the site. Specifically, this work is not intended to add new plugins like `doxylink`, `breathe`, or `MyST`. These will be added and tracked separately.
- Simplifying linking to content beyond that provided by built-in Sphinx functions for content in this repository. Specifically, I anticipate that we can make linking to the Python API significantly easier because everything will be done by Sphinx.
- Modify the theme. To the extent possible, we do not want to modify the theme of the site in this issue. This is a loose requirement; it may be easier to switch the theme to the PyData theme at this time, but I hope we can keep that out-of-scope.

As the necessary plugins will be changing in this switch, this issue supersedes #22 

<details><summary>Further context can be found by expanding this section.</summary>

In 2018, we re-designed the Cantera documentation and website using Nikola as the HTML generator for the bulk of the content, and retained Sphinx to generate API documentation in Cantera/cantera. At the time, this was justified on the basis that Nikola would be better suited to generate purely static content.

However, our experience is that mixing Nikola and Sphinx is more challenging than helpful. Specifically, the custom plugins that we implemented to replace parts of Sphinx's functionality for linking to other sections and documentation are not very robust, and since they're not documented, making updates or even just learning how to use them in prose is difficult. On the other hand, Sphinx roles are well known features and are well documented, so using Sphinx throughout will allow us to eliminate these custom Nikola plugins.

Second, the last 4 years have seen the development of fairly advanced Sphinx themes targeting the PyData community, specifically https://github.com/pydata/pydata-sphinx-theme. By switching to Sphinx, we can use this theme and eliminate many of the theme customization we developed for Nikola, including the custom Bootstrap directives.

Finally, there is a new markup syntax, [MyST](https://myst-parser.readthedocs.io/en/latest/index.html) that offers the best of Markdown and reST mashed together. As they have native integration with Sphinx, we can easily use this markup by changing everything to Sphinx.
</details>

### LINKED ISSUE #211: Use MyST as the markup format
The [MyST markup format](https://myst-parser.readthedocs.io/en/latest/index.html) is a combination of the best parts of Markdown and RestructuredText. It can be installed as a Sphinx extension, so upon completion of #209 we shouldn't have much trouble to add this as a dependency. I think tackling this issue is best done after completing #210, to avoid conflicts with any content changes needed for that issue.

We should consider using `MyST-NB` so that the documentation pages (especially the tutorials) can be executed at build time. Content can still be written as Markdown files, but code markup sections become executable, which would be useful to smoke out errors in content due to API changes, etc.

## Action Items

- [x] Add dependency on the MyST parser extension
- [x] Update Sphinx configuration to use this extension
- [x] Use [sphinx-design](https://sphinx-design.readthedocs.io/en/pydata-theme/get_started.html) for native Bootstrap theme support (ADDED 18-JUN-2022, please pull the updated branch :smile:)
- [ ] Convert all prose content (tutorials, science, etc.) to use MyST markup

## Non-Goals

- Significant modifications or restructuring of content. This issue should focus on converting existing content to MyST format, not on adding, extending, modifying, or restructuring content.

### LINKED ISSUE #229: Allow access to latest documentation version through stable URL
As it is now, if I were wanting to link to the `ct.Reactor` documentation I would link to https://cantera.org/documentation/docs-2.6/sphinx/html/cython/zerodim.html#reactor. When Cantera 3.0 is released, the link will no longer point at the latest documentation. It would be nice if the latest version of the documentation were always accessible at a stable URL; for instance, with NumPy the latest version of documentation for `np.nan` would always be at https://numpy.org/doc/stable/reference/constants.html#numpy.NAN. 

I am proposing that the URL `https://cantera.org/documentation/stable/...` would currently host the 2.6 documentation, until the point at which 3.0 is released. At that point, the same URL would then point at the 3.0 documentation, and the 2.6 documentation would be moved to `https://cantera.org/documentation/2.6/...`. It would be especially helpful if there were a version selector on the documentation page in this case.

On a related note, it would also be nice if the URLs were shorter. Currently the python documentation is prefixed by `sphinx/html/cython` which is understandable, as it is informative of the underlying implementation, but I think simply using `python` as the prefix could be much more user friendly. Similarly, substituting `docs` for `documentation`. For example: `https://cantera.org/docs/stable/python/zerodim.html#reactor`.

I'd be curious to hear others thoughts on this. I'm not sure if there are any limitations on how this could break existing links.

speth:
**Changes**
- Move the "outer" non-versioned portion of the website from Nikola to Sphinx, providing a consistent visual appearance
  - Closes #209
- Convert pages from ReST to MyST Markdown format
  -  Closes #211
- Remove content that has been migrated into the versioned documentation
  - Closes https://github.com/Cantera/enhancements/issues/178
- Adjust paths to different pages to shorten URLs and to make it possible to create links to "stable" and "dev" docs without picking a specific version
  - Resolves #229.

**TODO**
- [x] Update CI configuration to run Sphinx

A version of the website built from this is available for exploration at https://testing.cantera.org. Since the new section layout is only currently available in the "dev" docs, this site has the "stable" docs replaced with the "dev" docs.


pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+  "--port",
+  "0",
+  "--open-browser",
+] }
+
+[tool.pdm.options]
+add = ["--no-self"]
+sync = ["--no-self"]
+install = ["--no-self"]
+update = ["--no-self"]
```
bryanwweber, line 47: `pdm` (at least as of 2.11) has an option to set up the project as an application. It's undocumented unless you start the project from scratch, but it goes:

```toml
[tool.pdm]
package-type = "application"
```
Then you can remove these `no-self` options.

bryanwweber:
@speth Do you mind if I force-push this branch to remove my work email from the commits? Just want to check so you know when to `reset --hard` 😄

ischoegl (commented):
Thank you @speth (and @bryanwweber) for all your work on this.

From my end, I had a brief look at [testing.cantera.org](https://testing.cantera.org/), but at the moment it still appears to be the Nikola theme? Or is it an exact clone in terms of appearance with the backend replaced?

bryanwweber:
> @speth Do you mind if I force-push this branch to remove my work email from the commits? Just want to check so you know when to `reset --hard` 😄

@speth all set, thank you!

speth:
> From my end, I had a brief look at [testing.cantera.org](https://testing.cantera.org/), but at the moment it still appears to be the Nikola theme? Or is it an exact clone in terms of appearance with the backend replaced?

Sorry, looks like I had messed something up and what I had deployed got clobbered. I think it should be working now (and shouldn't get overwritten unless something gets pushed to the `testing` branch).

source/conf.py
```diff
@@ -0,0 +1,154 @@
...
+# -- Project information -----------------------------------------------------
+
+project = "Cantera"
+copyright = "2023, Cantera Developers"
+author = "Cantera Developers"
+html_title = "Cantera"
+
+# The full version, including alpha/beta/rc tags
+version = "2.6"
+release = "2.6.0"
```
bryanwweber, line 31: These look out-of-date, unless they're read from somewhere else?

source/conf.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Print warnings into the built HTML files
+keep_warnings = True
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions: list[str] = [
```
bryanwweber, line 47:
```suggestion
extensions = [
```

source/conf.py
```diff
@@ -0,0 +1,154 @@
...
+root_doc = "index"
+
+# -- Internationalization ---------------------------------------------------
+language = "en"
+
+intersphinx_mapping = {
+    'stable': ('https://testing.cantera.org/stable', None),
+    'dev': ('https://testing.cantera.org/dev', None),
+    'ct30': ('https://testing.cantera.org/3.0/sphinx/html', None),
+    'ct26': ('https://testing.cantera.org/2.6/sphinx/html', None),
```
bryanwweber, line 81: These all still say `testing.cantera.org`?

source/conf.py
```diff
@@ -0,0 +1,154 @@
...
+html_static_path = ["_static"]
+html_logo = "_static/img/cantera-logo.png"
+html_favicon = "_static/img/favicon.png"
+html_css_files = ["css/custom.css"]
+html_js_files = ["js/copybutton.js"]
+html_sidebars: dict[str, list | list[str]] = {
+    "index": [],
+    "news/**": [
+          'ablog/postcard.html', 'ablog/recentposts.html',
+          'ablog/tagcloud.html']
```
bryanwweber, line 103: Are these valid links? The `ablog` prefix doesn't seem right, although these could be generated pages.

bryanwweber (changes requested):
A few things look fishy with the configuration, but all the content changes seem fine :smile:

source/conf.py
speth: Oh, I guess the copyright year needs to be updated. I believe the version info here doesn't actually get shown anywhere, and strictly speaking doesn't correspond to a specific Cantera version in any case.

source/conf.py
speth: Right, good catch.

source/conf.py
speth: These refer to templates within `ablog`: See https://ablog.readthedocs.io/en/stable/manual/templates-themes.html#sidebars and https://github.com/sunpy/ablog/tree/main/src/ablog/templates/ablog.

source/conf.py
bryanwweber: I think sphinx usually puts this in some metadata somewhere. You're right, though, that this doesn't correspond to a particular version. Maybe try deleting the versions and see if things break? If they can be deleted that seems like the best route

bryanwweber (approved):

